PR #91, opened 2022-12-20
Improved pending transaction tracking

KrzysztofLabus-RDX:


src/RadixDlt.NetworkGateway.PostgresIntegration/Models/PendingTransaction.cs
```diff
@@ -216,6 +216,8 @@ public void MarkAsCommitted(bool succeeded, DateTime timestamp)
     {
         Status = succeeded ? PendingTransactionStatus.CommittedSuccess : PendingTransactionStatus.CommittedFailure;
         CommitTimestamp = timestamp;
+        FailureReason = null;
```
dhedey, line 219: Hmm if it's a committed failure, we should ideally include the failure reason here?

dhedey (approved):
Approving - looks good, maybe one final fix regarding saving committed failure to pending transaction - although it's possibly not super important!

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/PendingTransaction.cs
KrzysztofLabus-RDX: Precisely. This should be renamed to "most recent transient error/failure reason"

PR #108, opened 2023-01-13
ValueBytes

KrzysztofLabus-RDX:


.editorconfig
```diff
@@ -283,6 +283,7 @@ dotnet_diagnostic.SA1649.severity = none # SA1649FileNameMustMatchTypeName - it
 dotnet_diagnostic.SA0001.severity = none # SA0001XmlCommentAnalysisDisabled
 dotnet_diagnostic.SA1101.severity = none
 dotnet_diagnostic.SA1124.severity = none
+dotnet_diagnostic.SA1129.severity = none
```
dhedey, line 286: Can we add a comment about what it is we're disabling?

dhedey (approved):
Nice change. Approve of this -- missing the EqualityComparer was an easy bug waiting to happen somewhere!

.editorconfig
KrzysztofLabus-RDX: Added

PR #109, opened 2023-01-17
CoreAPI Upgrade

KrzysztofLabus-RDX:
**Draft:** awaits for PR in Babylon-node


generation/regenerate-core-api.sh
```diff
@@ -2,6 +2,13 @@
 
 set -e
 
+# !!!WARNING!!!
+# OAS file must be manually changed so that:
```
dhedey, line 6: You might want to try out a python script like we have in the node - and perform these changes automatically? 👍

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/ReferencedEntity.cs
```diff
@@ -91,7 +91,7 @@ internal record ReferencedEntity(string IdHex, CoreModel.EntityType Type, long S
 
     public long DatabaseGlobalAncestorId => GetDatabaseEntity().GlobalAncestorId ?? throw new InvalidOperationException("GlobalAncestorId not set, probably global entity or incorrectly configured one.");
 
-    public bool CanBeOwner => Type is CoreModel.EntityType.Component or CoreModel.EntityType.ResourceManager or CoreModel.EntityType.KeyValueStore;
+    public bool CanBeOwner => Type is CoreModel.EntityType.Component or CoreModel.EntityType.ResourceManager or CoreModel.EntityType.KeyValueStore or CoreModel.EntityType.Package;
```
dhedey, line 94: It might be easier to just assume anything can be an owner :)

dhedey (approved):

generation/regenerate-core-api.sh
KrzysztofLabus-RDX: I'll do it once we move SDKs into separate repo. Then I'll do my best to make sure C#, Java, TypeScript models and SDKs are generated using more or less same tooling in a consistent manner.

PR #110, opened 2023-01-17
Dropped Ledger Target State

KrzysztofLabus-RDX:


dhedey (approved):
Looks like a pretty easy removal in the end? 👍  - we might need to just slightly adapt our metrics?

Was it used by the Core API health check in the Gateway API? Or did I misremember?

PR #112, opened 2023-01-25
feat: merged interactive devenv with other devenvs

KrzysztofLabus-RDX:
As of right now instead of running `docker-compose up` for local development purposes you're expected to run `./development/run-only-network-gateway-resources.sh` or other of those preconfigured scripts.

This PR solves several problems:

1. We reduce overall number of different "environments" our apps are run in by one.
2. Feature-proofing: it is expected NG will depend on few extra resources (think: Redis, RabbitMQ etc.) so we eliminate the need to configure them twice.
3. The exact version numbers of our external dependencies (including Node) can be now tracked as part of our VCS, i.e. our ephemeral environments can rely on `.env/.template.env` to figure out what version of postgres / node should be deployed.


deployment/.template.env
```diff
@@ -3,6 +3,17 @@
...
+########################################################
+# DOCKER IMAGE VERSION TAGS & DEVELOPMENT DEPENDENCIES #
+########################################################
+
+FULLNODE_DOCKER_TAG=2.0.0 # TODO not published yet
+POSTGRES_DB_DOCKER_TAG=14.3-alpine
+NETWORK_GATEWAY_DOCKER_TAG=2.0.0 # TODO not published yet
+
+# Git commit version of most recent Node compatible with this Network Gateway (only used in CI/CD environments)
+FULLNODE_VERSION=cf1fc5657e42f3f6239c10555ff847cc856f40a9
```
dhedey, line 15: This is in the `.template.env` file, not the `.env` file - so I think this is wrong.

Each developer and CI should have all of these automatically update when they pull in the latest version.

I would suggest therefore that these live in a `versions.env` file at the root, and that this env file is loaded by the scripts automatically...

And we leave this `.template.env` for defaults to the _changes/overrides_ that users may wish to make themselves.

dhedey: Although I don't know if docker compose can handle that.

It might be that we need to create a `.env` file on the fly by combining the `/versions.env` with a `/deployment/settings.env` file to create a temporary `.env` file which is then used by docker compose?

dhedey: I imagine the Gateway ephemeral environment stuff will also want to read these, so having them in a separate `/versions.env` file will help them read them.

dhedey (commented):
Nice thinking, although I'm not sure it's quite right, as discussed in the comments

deployment/.template.env
KrzysztofLabus-RDX: > Each developer and CI should have all of these automatically update when they pull in the latest version.

That's precisely what we want NOT to happen, we want to control this Node version manually (just as say NuGet packages)

> Although I don't know if docker compose can handle that.

It can't, this is why I've decided to place all of that in the `.template.env` file which I wanted our Ephemeral envs to use. Why? To reduce number of extra scripts/steps to execute in order to just start all of that. 
Have I essentially piggybacked some CI-specific logic into Docker-specific file? Yes, and I think as long as it is literally single variable this should be fine. Obviously I could move this single variable (`FULLNODE_VERSION`) into `development/ci.env` or something.

Theoklitos:
I would generally prefer if the version was stored in `/versions.env` or `development/ci.env`, though I can still work with wherever its located.

More importantly, the ephemeral jobs work with docker tags, not git commits.

Using a git commit would basically force us to checkout, build and publish the node image on every PR push. This would take a lot of time.

Perhaps we can specify both? The git commit _and_ the docker tag (if available).

KrzysztofLabus-RDX:
@dhedey  @Theoklitos  Please review again.

PR #123, opened 2023-01-31
CoreApi upgrade

KrzysztofLabus-RDX:


src/RadixDlt.NetworkGateway.GatewayApi/Services/NetworkConfigurationProvider.cs
```diff
@@ -77,10 +77,12 @@ public interface INetworkConfigurationProvider : INetworkAddressConfigProvider
 {
     Task Initialize(ICapturedConfigProvider capturedConfigProvider, CancellationToken token);
 
+    int GetNetworkId();
```
PawelPawelec-RDX, line 80: toolkit library uses byte type, any reason for it to be int then?

dhedey: For Krz's benefit - this network id is a byte - although in Core API it's represented as an int32 because OpenAPI can't do bytes :P. It does have a min:0 / max: 255 on though 👍

KrzysztofLabus-RDX: You're obviously right, fixed.

PawelPawelec-RDX (approved):

dhedey (approved):

PR #129, opened 2023-02-01
Store global address in database

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.Abstractions/RadixGlobalAddress.cs
```diff
@@ -0,0 +1,87 @@
...
+    private readonly string _address;
+
+    private RadixGlobalAddress(string address)
+    {
+        _address = address;
+    }
+
+    public static implicit operator string(RadixGlobalAddress ra) => ra._address;
+
+    public static implicit operator RadixGlobalAddress(string text) => new RadixGlobalAddress(text);
```
KrzysztofLabus-RDX, line 81: Just thinking out loud: isn't it too easy to mix `EntityGlobalAddress` with other types when we use implicit casting and no guard checks whatsoever? I mean... having method like `MyMethod(EntityId eid, EntityGlobalAddress ga)` we can still call it as `.MyMethod("abc_111111", "abcdef")` or `.MyMethod("abcdef", "abc_111")`.

1. Maybe we should use explicit casting when converting from string to wrapper type?
2. Maybe we could add some super lightweight sanity checks when creating wrapper types? Think `input.All(c => c  >= 'a' and <= 'f' or >= '0' <= '9')` for `EntityId` and `input.Contains('_') && input.All(c => c >= 'a' and <= 'z' or c >= '0' and <= '9')`?
3. Maybe those checks could be only used for implicit casts?

KrzysztofLabus-RDX (changes requested):
Awesome to see so many lines removed! Great job, just few minor aspects to discuss.

src/RadixDlt.NetworkGateway.Abstractions/RadixGlobalAddress.cs
```diff
@@ -0,0 +1,87 @@
...
+ * Work. You assume all risks associated with Your use of the Work and the exercise of
+ * permissions under this License.
+ */
+
+using System.Diagnostics;
+
+namespace RadixDlt.NetworkGateway.Abstractions;
+
+[DebuggerDisplay("{ToString()}")]
+public sealed record RadixGlobalAddress
```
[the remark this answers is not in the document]
PawelPawelec-RDX: yes exactly, wanted to keep it consistent. I'll just rename RadixGlobalAddress => GlobalAddress and I'll take care of RadixAddress => EntityId renaming in my next task because they overlap.

src/RadixDlt.NetworkGateway.Abstractions/RadixGlobalAddress.cs
PawelPawelec-RDX: **Ad. 1 and 2)** I think, you're right. Even when migrating the code I made a mistake once and simply passed invalid string. I think your idea to create explicit cast and add sanity checks there is a good idea.

Do you think we should create separate sanity check logic or reuse some existing one i.e https://github.com/radixdlt/babylon-gateway/blob/4f34959dfb43a5200847cf7bae8a323bbefdec92/src/RadixDlt.NetworkGateway.Abstractions/Addressing/Bech32Codec.cs#L248 ? 

Logic there is pretty simple so it shouldn't affect performance that much and we'd not duplicate validation logic, what do you think?

**Ad. 3)** to be honest I'd probably leave explicit cast only.

KrzysztofLabus-RDX: OK, so explicit cast from string to wrapper type, implicit cast from wrapper type to string and no extra checks? Sounds good to me.

PawelPawelec-RDX: what about reusing that validation method?

KrzysztofLabus-RDX: It is potentially bit heavy, I'd say not to use it when data comes from trusted source (our internal databases etc.), i.e. let's skip that.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/EntityStateQuerier.cs
```diff
@@ -217,7 +218,7 @@ public async Task<GatewayModel.EntityDetailsResponse> EntityDetailsSnapshot(Deco
 
                 details = new GatewayModel.EntityDetailsResponseDetails(new GatewayModel.EntityDetailsResponseComponentDetails(
                     discriminator: GatewayModel.EntityDetailsResponseDetailsType.Component,
-                    packageAddress: package.BuildHrpGlobalAddress(_networkConfigurationProvider.GetHrpDefinition()),
+                    packageAddress: package.GlobalAddress!,
```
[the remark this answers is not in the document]
PawelPawelec-RDX: got rid of it, it's not really needed anymore I guess after changing to `record struct`

KrzysztofLabus-RDX (approved):

PR #139, opened 2023-02-03
fix viewmodel type in viewmodel after changing global address in db to string

PawelPawelec-RDX:


KrzysztofLabus-RDX (approved):
Approving but shouldn't we immediately switch to `GlobalAddress` rather `string`?

PawelPawelec-RDX:
> Approving but shouldn't we immediately switch to `GlobalAddress` rather `string`?

@KrzysztofLabus-RDX good idea, created custom mapper.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/EntityStateQuerier.cs
```diff
@@ -86,9 +86,9 @@ internal class EntityStateQuerier : IEntityStateQuerier
 {
     private record MetadataViewModel(long EntityId, string[] Keys, string[] Values, int TotalCount);
 
-    private record FungiblesViewModel(byte[] ResourceEntityGlobalAddress, string Balance, int TotalCount);
+    private record FungiblesViewModel(GlobalAddress? ResourceEntityGlobalAddress, string Balance, int TotalCount);
```
KrzysztofLabus-RDX, line 89: Those can't be null?

KrzysztofLabus-RDX (changes requested):

KrzysztofLabus-RDX (approved):

PR #141, opened 2023-02-07
DO-1167-fetch docker credentials from aws

kofrdx:
Store Dockerhub credentials in AWS

This is analog to babylon-node and babylon-nginx

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -66,6 +66,10 @@ jobs:
   publish-docker:
     runs-on: ubuntu-22.04
     environment: release
+    permissions:
+      id-token: write
+      contents: read
+      pull-requests: write
     steps:
       - name: Dump context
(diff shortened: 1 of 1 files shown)


.github/workflows/releases.yml
```diff
@@ -80,11 +84,24 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
 
-      - name: Login to Dockerhub (release)
-        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a
+      - name: "Configure AWS credentials"
+        uses: aws-actions/configure-aws-credentials@v1-node16
```
shambupujar, line 88: It would be good if you can change this to SHA reference as tags can be modified on source repo to point different commit

kofrdx: updated

PR #153, opened 2023-02-10
feat: Added CHANGELOG.md and known bugs section to README.md

KrzysztofLabus-RDX:


CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
+### 0.1.1
```
PawelPawelec-RDX, line 1: is Betanet v2 just internal release name? or do we want to include it here.

KrzysztofLabus-RDX: I guess it should be only considered internal? We can always update this changelog post-factum.

PawelPawelec-RDX (approved):

PR #164, opened 2023-02-15
Transaction Stream

KrzysztofLabus-RDX:


src/RadixDlt.NetworkGateway.PostgresIntegration/Services/TransactionQuerier.cs
```diff
@@ -222,34 +215,43 @@ private async Task<DetailsLookupResult> GetTransactionWithDetails(long stateVers
                 .ToListAsync(token);
         }
 
-        return MapToGatewayAccountTransactionWithDetails(transaction, rawTransaction, referencedEntities);
+        return MapToGatewayAccountTransactionWithDetails(transaction, referencedEntities);
     }
 
-    private GatewayModel.CommittedTransactionInfo MapToGatewayAccountTransaction(UserLedgerTransaction ult)
+    private GatewayModel.CommittedTransactionInfo MapToGatewayAccountTransaction(LedgerTransaction lt)
```
PawelPawelec-RDX, line 221: can we replace these abbreviations here with meaningful names?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -305,6 +280,31 @@ paths:
         "4XX":
           $ref: "#/components/responses/ErrorResponse"
 
+  "/stream/transactions":
+    post:
+      operationId: StreamTransactions
+      summary: Get Transactions Stream
+      description: |
+        Returns transactions which have been committed to the ledger.
+        The returned response is in a paginated format, ordered by most recently committed.
```
PawelPawelec-RDX, line 289: that isn't always true. There's a parameter which can control ordering.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1182,6 +1161,46 @@ components:
...
+            transaction_type:
+              type: string
+              enum:
+                - user
+                - epoch_change
+                - all
+            from_ledger_state:
+              $ref: "#/components/schemas/LedgerStateSelector"
+            order:
+              type: string
```
PawelPawelec-RDX, line 1183: should we mention what's the default ordering?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1182,6 +1161,46 @@ components:
...
+    # /stream/* Request & Response Types
+    #
+
+    StreamTransactionsRequest:
+      allOf:
+        - $ref: "#/components/schemas/AtLedgerStateMixin"
+        - $ref: "#/components/schemas/CursorLimitMixin"
+        - type: object
+          properties:
+            transaction_type:
```
PawelPawelec-RDX, line 1174: probably I'm just missing some information but that wasn't part of endpoint specs. I'm asking because we're adding `epoch_change` as transaction type which isn't really a transaction type?

If i understand things right all transaction types can result in new epoch. Exposing such type to outer world might cause confusion that it's actual transaction type and it might be also confusing that user asked for transaction_type `epoch_change` and we're returning him all sort of different transaction types.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -693,6 +693,18 @@ components:
...
           format: int64
 
+    CursorLimitMixin:
+      type: object
+      properties:
+        cursor:
+          description: This cursor allows forward pagination, by providing the cursor from the previous request.
+          type: string
+          nullable: true
+        limit_per_page:
```
PawelPawelec-RDX, line 703: should we specify upper limit in specs maybe and default value if not provided?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1182,6 +1161,46 @@ components:
         message_hex:
           $ref: "#/components/schemas/HexString"
 
+    #
+    # /stream/* Request & Response Types
+    #
+
+    StreamTransactionsRequest:
```
PawelPawelec-RDX, line 1168: I know that sbor encoding is not possible to address now but what with other mixins:

- raw_payload
- Parts of the receipt (state_changes, fee_summary, events)

is it something we're skipping or delaying?

PawelPawelec-RDX (commented):
what about e2e tests for that endpoint? I know it's part of other repository but we should plan adding some in scope of that task.

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/LedgerTransaction.cs
```diff
@@ -140,6 +140,18 @@ internal abstract class LedgerTransaction
     /// </summary>
     [Column("normalized_round_timestamp")]
     public DateTime NormalizedRoundTimestamp { get; set; }
+
+    [Column("filter_type")]
+    public LedgerTransactionFilterType? FilterType { get; set; }
```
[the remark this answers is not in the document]
KrzysztofLabus-RDX: Added some docs to make it slightly clearer.

src/RadixDlt.NetworkGateway.GatewayApi/Handlers/DefaultTransactionHandler.cs
```diff
@@ -75,6 +75,8 @@ namespace RadixDlt.NetworkGateway.GatewayApi.Handlers;
 
 internal class DefaultTransactionHandler : ITransactionHandler
 {
+    private const int DefaultPageLimit = 10; // TODO make it configurable
```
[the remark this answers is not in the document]
KrzysztofLabus-RDX: There's like 5 or 10 of them. I'd rather prefer to handle them in a separate PR, created tracking ticket though NG-288

src/RadixDlt.NetworkGateway.GatewayApi/Handlers/DefaultTransactionHandler.cs
```diff
@@ -176,4 +158,28 @@ public async Task<GatewayModel.TransactionSubmitResponse> Submit(GatewayModel.Tr
...
+    {
+        var atLedgerState = await _ledgerStateQuerier.GetValidLedgerStateForReadRequest(request.AtLedgerState, token);
+        var fromLedgerState = await _ledgerStateQuerier.GetValidLedgerStateForReadForwardRequest(request.FromLedgerState, token) ?? atLedgerState;
+
+        var transactionsPageRequest = new TransactionStreamPageRequest(
+            FromStateVersion: fromLedgerState.StateVersion,
+            Cursor: GatewayModel.LedgerTransactionsCursor.FromCursorString(request.Cursor),
+            PageSize: request.LimitPerPage ?? DefaultPageLimit,
+            AscendingOrder: request.Order == GatewayModel.StreamTransactionsRequest.OrderEnum.Asc,
+            UserTransactionsOnly: request.TransactionType is null or GatewayModel.StreamTransactionsRequest.TransactionTypeEnum.User,
```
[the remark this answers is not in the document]
KrzysztofLabus-RDX: I know we're actually using GatewayModel in our service layer today but that's a bad design where API surface models leak into the deepest part of services. And it's not only models but also logic - like the fact that `User` is essentially a default value. I don't want to create even more tech debt.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/TransactionQuerier.cs
```diff
@@ -157,47 +148,52 @@ public async Task<ICollection<StatusLookupResult>> LookupPendingTransactionsByIn
...
-                .OrderBy(at => at.StateVersion)
-                .Take(request.PageSize + 1)
-                .Select(at => at.StateVersion)
-                .ToListAsync(token);
+        var query = _dbContext.LedgerTransactions
+            .Where(lt => lt.StateVersion <= atLedgerState.StateVersion);
+
+        if (request.AscendingOrder)
+        {
+            query = query.Where(lt => lt.StateVersion >= stateVersionBoundary)
```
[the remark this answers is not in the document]
KrzysztofLabus-RDX: Because with ascending order `next_cursor` is going to point to the higher state version while with regular descending order it is going to be lower one.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/TransactionQuerier.cs
```diff
@@ -157,47 +148,52 @@ public async Task<ICollection<StatusLookupResult>> LookupPendingTransactionsByIn
...
+        }
+
+        return await query.Take(request.PageSize + 1)
+            .Select(at => at.StateVersion)
+            .ToListAsync(token);
     }
 
     private async Task<List<GatewayModel.CommittedTransactionInfo>> GetTransactions(List<long> transactionStateVersions, CancellationToken token)
     {
         var transactions = await _dbContext.LedgerTransactions
```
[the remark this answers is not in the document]
KrzysztofLabus-RDX: `.AsNoTracking()` is configured on `dbContext` level.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
KrzysztofLabus-RDX: Good catch!

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
KrzysztofLabus-RDX: Good catch!

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
KrzysztofLabus-RDX: I'll try to find out a different name to reduce confusion.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
KrzysztofLabus-RDX: This is definitely outside the scope of this PR and I'm personally not even sure how handle that.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/TransactionQuerier.cs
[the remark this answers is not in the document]
PawelPawelec-RDX: ok so I as susspected.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: hmm you're right, we return meaningful error response, so should be ok. Is that documented anywhere or returned from some api "configuration alike" endpoint? or api user has to check it manually what is the limit?

src/RadixDlt.NetworkGateway.GatewayApi/Handlers/DefaultTransactionHandler.cs
```diff
@@ -176,4 +161,36 @@ public async Task<GatewayModel.TransactionSubmitResponse> Submit(GatewayModel.Tr
...
+        var atLedgerState = await _ledgerStateQuerier.GetValidLedgerStateForReadRequest(request.AtLedgerState, token);
+        var fromLedgerState = await _ledgerStateQuerier.GetValidLedgerStateForReadForwardRequest(request.FromLedgerState, token) ?? atLedgerState;
+
+        var kindFilter = request.KindFilter switch
+        {
+            GatewayModel.StreamTransactionsRequest.KindFilterEnum.All => LedgerTransactionKindFilter.AllAnnotated,
+            GatewayModel.StreamTransactionsRequest.KindFilterEnum.User => LedgerTransactionKindFilter.UserOnly,
+            GatewayModel.StreamTransactionsRequest.KindFilterEnum.EpochChange => LedgerTransactionKindFilter.EpochChangeOnly,
+            null => LedgerTransactionKindFilter.UserOnly,
+            _ => throw new UnreachableException(),
```
PawelPawelec-RDX, line 176: could we pass argument value that wasn't expected as exception message? That would improve observability and debugging by a lot. I guess it's worth revisiting and fixing all UnreachableException (we have just 7 of them should be quick fix).

PawelPawelec-RDX (approved):
Approving just minor proposals/questions feel free to ignore and merge.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
KrzysztofLabus-RDX: What the limit is isn't exposed anywhere and that is a OK thing I believe. It is optionally up to the NG owner to document that in some handcrafted docs (think: user guide)

src/RadixDlt.NetworkGateway.GatewayApi/Handlers/DefaultTransactionHandler.cs
KrzysztofLabus-RDX: I've personally mixed feelings about `UnreachableException` I used here and there over good-old `ArgumentOutOfRangeException` but let's discuss that outside of this PR.

PR #169, opened 2023-02-16
Use single solution build within multiple docker project builds

krzlabrdx:


krzlabrdx:
Closed due to inactivity.

PR #170, opened 2023-02-17
use ret library for static validation

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.GatewayApi/Configuration/EndpointOptions.cs
```diff
@@ -76,6 +76,9 @@ public sealed class EndpointOptions
 
     [ConfigurationKeyName("RequestTimeout")]
     public TimeSpan RequestTimeout { get; set; } = TimeSpan.FromSeconds(10);
+
+    [ConfigurationKeyName("SubmitTransactionTimeout")]
+    public TimeSpan SubmitTransactionTimeout { get; set; } = TimeSpan.FromSeconds(3);
```
KrzysztofLabus-RDX, line 81: IMO this is something that should be placed under some "core api integration options" rather "endpoint options" but I guess this is something we could remodel later if you want to.

Please make sure to account for this new option in our changelog.

KrzysztofLabus-RDX (changes requested):

src/RadixDlt.NetworkGateway.GatewayApi/Configuration/EndpointOptions.cs
PawelPawelec-RDX: ok extracted that and added info to changelog.

KrzysztofLabus-RDX (approved):

PR #173, opened 2023-02-22
Order resources by their most recent update state version

KrzysztofLabus-RDX:


src/RadixDlt.NetworkGateway.PostgresIntegration/Models/EntityResourceAggregateHistory.cs
```diff
@@ -84,12 +86,41 @@ internal class EntityResourceAggregateHistory
...
+    [Column("non_fungible_resource_significant_update_state_versions")]
+    public List<long> NonFungibleResourceSignificantUpdateStateVersions { get; set; }
+
+    public void MarkAsCloneOfExistingRecord()
+    {
+        _fungibleResourceEntityIdsOrg = FungibleResourceEntityIds.ToArray();
+        _nonFungibleResourceEntityIdsOrg = NonFungibleResourceEntityIds.ToArray();
+    }
+
+    public bool ShouldBePersisted()
```
PawelPawelec-RDX, line 104: that piece of code would be easily testable by unit tests, maybe we should add some?

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/EntityResourceAggregateHistory.cs
```diff
@@ -84,12 +86,41 @@ internal class EntityResourceAggregateHistory
     [Column("fungible_resource_entity_ids")]
     public List<long> FungibleResourceEntityIds { get; set; }
 
-    [Column("fungible_resource_last_update_state_versions")]
-    public List<long> FungibleResourceLastUpdateStateVersions { get; set; }
+    [Column("fungible_resource_significant_update_state_versions")]
```
PawelPawelec-RDX, line 89: I could come with better name for these columns so I think we have to leave them as it is. But if we can't find better name, what do you think about adding comment here to explaining what it is about?

PawelPawelec-RDX (approved):
just minor comments, feel free to merge.

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/EntityResourceAggregateHistory.cs
KrzysztofLabus-RDX: Refactored EntityResourceAggregateHistory quite heavily.

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/EntityResourceAggregateHistory.cs
KrzysztofLabus-RDX: Refactored EntityResourceAggregateHistory quite heavily.

PawelPawelec-RDX (approved):

PR #187, opened 2023-03-01
Merge entity id and global_address in OAS

KrzysztofLabus-RDX:


PawelPawelec-RDX (approved):
looks good, not sure if we shouldn't put info about that in changelog.

PR #189, opened 2023-03-02
update network node dependency.

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1072,16 +1072,13 @@ components:
...
             - clock
             - ecdsa_secp256k1
             - eddsa_ed25519
             - xrd
           properties:
-            account_package:
-              $ref: "#/components/schemas/EntityAddress"
             faucet:
               $ref: "#/components/schemas/ComponentAddress"
             epoch_manager:
```
KrzysztofLabus-RDX, line 1084: I'm quite sure we should also change NFID Type `Number` to `Integer`.

KrzysztofLabus-RDX (changes requested):

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: done.

KrzysztofLabus-RDX (approved):

PR #191, opened 2023-03-02
chore: reusable docker build action

kofrdx:


kofrdx:
Waiting for fix of the develop branch to the application build. After that I can test the dockerhub release.

.github/workflows/releases.yml
```diff
@@ -58,92 +58,120 @@ jobs:
...
+      image_name: "babylon-ng-database-migrations"
+      tag: ${{ needs.setup-tags.outputs.database-migrations-tag }}
+      context: "."
+      dockerfile: "./apps/DatabaseMigrations/Dockerfile"
+      platforms: "linux/amd64"
+      environment: release
+      cache_tag_suffix: "release"
+      enable_dockerhub: "true"
+      restore_artifact: "true"
+      artifact_location: "src/DataAggregator/appsettings.json"
```
shambupujar, line 122: May need to understand , Why do we need this. This file is available within context ?

.github/workflows/releases.yml
```diff
@@ -58,92 +58,120 @@ jobs:
...
-          password: ${{env.DOCKERHUB_TOKEN}}
-
       - name: Setup tags for docker image
         id: setup_tags
         uses: ./.github/actions/set-variables
         with:
           github_event_name: ${{ github.event_name }}
           github_action_name: ${{ github.event.action}}
-
       - name: Setup application version
```
kofrdx, line 80: @shambupujar  This is where the file you mentioned is created. I assume its important for the application to know its own release version. I create this file inside a different job, so I need to fetch it via artifact_restore. After that its available in the docker build context and gets copied.

.github/workflows/releases.yml
kofrdx: Check here: https://github.com/radixdlt/babylon-gateway/pull/191/files#r1131132885
Also the comment below to for more explanation

.github/workflows/releases.yml
```diff
@@ -58,92 +58,120 @@ jobs:
...
         run: |
           echo "$( jq '.+{DataAggregatorVersion:"${{ steps.get_release.outputs.tag_name }}"}' apps/DataAggregator/appsettings.json )" > src/DataAggregator/appsettings.json
           echo "$( jq '.+{DatabaseMigrationsVersion:"${{ steps.get_release.outputs.tag_name }}"}' apps/DatabaseMigrations/appsettings.json )" > src/DatabaseMigrations/appsettings.json
           echo "$( jq '.+{GatewayApiVersion:"${{ steps.get_release.outputs.tag_name }}"}' apps/GatewayApi/appsettings.json )" > src/GatewayAPI/appsettings.json
```
kofrdx, line 84: @shambupujar This is where the file you mentioned is created. I assume its important for the application to know its own release version. I create this file inside a different job, so I need to fetch it via artifact_restore. After that its available in the docker build context and gets copied.

.github/workflows/releases.yml
kofrdx: And the cache_tag_suffix is entirely optional. Could also be the same cache image/tag as for the pr. I though about seperating them for making the cache invalidation more predictable.

shambupujar: Ah okay. in that case , this should be database migration one
https://github.com/radixdlt/babylon-gateway/pull/191/files#diff-a62ff2d992534310d441210a2a8b31711c7e515c326a770d8b31f54a642d3f46R83

PR #208, opened 2023-03-17
Reworked /state/non-fungible/* endpoints

KrzysztofLabus-RDX:


src/RadixDlt.NetworkGateway.GatewayApi/Validators/StateNonFungibleDetailsRequestValidator.cs
```diff
@@ -69,16 +69,25 @@
...
 
-        RuleFor(x => x.NonFungibleId)
-            .NotEmpty();
+        RuleFor(x => x.NonFungibleIds)
+            .NotEmpty()
+            .DependentRules(() =>
+            {
+                RuleFor(x => x.NonFungibleIds.Count)
+                    .GreaterThan(0)
+                    .LessThan(endpointOptionsSnapshot.Value.DefaultPageSize);
```
PawelPawelec-RDX, line 86: hmm I guess that property was designed to hold default page size if value wasn't provided, not to hold upper limit. 

Guess we should use `MaxPageSize` ? I think it'd also make sense to change default value of MaxPageSize to be higher than default :D currently it's 30 and Default is 100 so if user wouldn't specify anything he'd get 100 items but he can't ask explicitly for more than 30.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1738,74 +1738,86 @@ components:
...
           properties:
-            address:
-              $ref: "#/components/schemas/EntityAddress"
+            resource_address:
+              $ref: "#/components/schemas/ResourceAddress"
             non_fungible_ids:
               $ref: "#/components/schemas/NonFungibleIdsCollection"
 
-    NonFungibleDataRequest: # TODO rename to NonFungibleIdDataRequest
+    StateNonFungibleDetailsRequest:
```
PawelPawelec-RDX, line 1769: If we allow querying by multiple non_fungible_ids shouldn't it be named `StateNonFungible Ids DetailsRequest` ?

PawelPawelec-RDX (changes requested):

src/RadixDlt.NetworkGateway.GatewayApi/Validators/StateNonFungibleDetailsRequestValidator.cs
KrzysztofLabus-RDX: You're obviously right about both of those issues. Fixed.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/EntityStateQuerier.cs
```diff
@@ -392,28 +391,34 @@ LIMIT 1
...
+                entityId = entity.Id,
+                nonFungibleIds = nonFungibleIds,
+            },
             cancellationToken: token);
 
-        var data = await _dbContext.Database.GetDbConnection().QueryFirstOrDefaultAsync<NonFungibleIdDataViewModel>(cd);
+        var items = new List<GatewayModel.StateNonFungibleDetailsResponseItem>();
 
-        if (data == null || data.IsDeleted)
+        foreach (var vm in await _dbContext.Database.GetDbConnection().QueryAsync<NonFungibleIdDataViewModel>(cd))
```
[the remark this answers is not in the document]
KrzysztofLabus-RDX: I disagree, here's my reasoning: `foreach (var vm in await _dbContext.Database.GetDbConnection().QueryAsync<...>(cd))` is already used several times within this very file for slightly more complex operations and my brain's desire for repetition and patterns in general must be fulfilled. :)

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
KrzysztofLabus-RDX: Not really, not as far as I can tell. Actually here's some quick reading I've found on "nouns of nouns" - https://english.stackexchange.com/questions/255433/how-to-say-a-case-of-multiple-of-multiples-correctly

PawelPawelec-RDX (approved):

PR #225, opened 2023-03-28
Avoid mardunet deployment if ci.env is unchanged

ghost:


.github/workflows/main.yml
```diff
@@ -112,10 +112,17 @@ jobs:
...
+        id: changed-files
+        uses: tj-actions/changed-files@db5dd7c176cf59a19ef6561bf1936f059dee4b74
+        with:
+          files: |
+            deployment/ci.env
       - name: Trigger deployment event ${{ github.ref }}
         env:
           NAMESPACE: "ng-mardunet"
           EVENT_TYPE: "ng_babylon_mardunet"
+        if: steps.changed-files.outputs.any_changed == 'true'
```
shambupujar, line 125: You need to still deploy , but only gateway. The value on the respository dispatch should change and so then the workflows on radixdlt-iac needs to be udpated

shambupujar (approved):

PR #227, opened 2023-03-30
return unboxed transaction raw hex.

PawelPawelec-RDX:


src/RadixDlt.CoreApiSdk/Model/LedgerTransaction.cs
```diff
@@ -78,4 +78,19 @@ public byte[] GetPayloadBytes()
             _ => throw new UnreachableException($"Didn't expect {this.GetType().Name} type"),
         };
     }
+
+    // TODO:
+    // This is rcnet quick fix for returning transaction hex that's not boxed in user/validator/system enum by node.
+    // Boxed payload is not supported by the toolkit and client's can't actually do anything about it.
+    // Needs revisiting.
+    public byte[] GetUnboxedPayloadBytes()
```
dhedey, line 86: Might suggest "Unwrapped" rather than "Unboxed" because boxing has a special meaning in C#/programming, but otherwise good!

src/RadixDlt.CoreApiSdk/Model/LedgerTransaction.cs
```diff
@@ -78,4 +78,19 @@ public byte[] GetPayloadBytes()
             _ => throw new UnreachableException($"Didn't expect {this.GetType().Name} type"),
         };
     }
+
+    // TODO:
+    // This is rcnet quick fix for returning transaction hex that's not boxed in user/validator/system enum by node.
+    // Boxed payload is not supported by the toolkit and client's can't actually do anything about it.
+    // Needs revisiting.
```
dhedey, line 85: Yeah - this makes sense to revisit.
TBH I might suggest that we don't even return the ledger transaction, and let the Gateway create it itself on demand if it wants it, from the discriminator and sub-payload - or add it as an opt-in include.

dhedey (approved):
Approved.

A couple of minor thoughts - feel free to ignore them!

PR #245, opened 2023-04-14
Feature/explicit metadata

KrzysztofLabus-RDX:


src/RadixDlt.NetworkGateway.PostgresIntegration/Services/EntityStateQuerier.cs
```diff
@@ -623,6 +639,60 @@ LIMIT 1
...
+            foreach (var key in keys)
+            {
+                entityIds.Add(entityId);
+                metadataKeys.Add(key);
+            }
+        }
+
+        var metadataHistory = await _dbContext.EntityMetadataHistory
+            .FromSqlInterpolated(@$"
+WITH variables (entity_id, metadata_key) AS (
```
PawelPawelec-RDX, line 659: I guess we want some deterministic order of results returned from that SQL and later endpoint. Would probably make sense to order by `entity_id` then by `keys`?

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/EntityStateQuerier.cs
```diff
@@ -573,22 +583,25 @@ LIMIT 1
...
-        a.val AS entity_id,
-        a.ord AS ord
-     FROM UNNEST(@entityIds) WITH ORDINALITY AS a(val, ord)
+WITH variables (entity_id) AS (
+    SELECT UNNEST(@entityIds)
+),
+metadata_slices AS (
+    SELECT variables.entity_id, emah.metadata_slice, emah.metadata_total_count
+    FROM variables
+    INNER JOIN LATERAL (
```
PawelPawelec-RDX, line 592: i'm not sure if we're not losing ordering after that join, saw that you're applying that schema with ordinality in next select statement but isn't that too late?

src/RadixDlt.NetworkGateway.PostgresIntegration/CommonDbContext.cs
```diff
@@ -244,6 +246,12 @@ private static void HookupHistory(ModelBuilder modelBuilder)
         modelBuilder.Entity<EntityMetadataHistory>()
             .HasIndex(e => new { e.EntityId, e.FromStateVersion });
 
+        modelBuilder.Entity<EntityMetadataHistory>()
```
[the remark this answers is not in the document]
KrzysztofLabus-RDX: Removed unnecessary index. Good catch!

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/EntityMetadataAggregateHistory.cs
```diff
@@ -0,0 +1,86 @@
...
+    public long Id { get; set; }
+
+    [Column("from_state_version")]
+    public long FromStateVersion { get; set; }
+
+    [Column("entity_id")]
+    public long EntityId { get; set; }
+
+    [Column("metadata")]
+    public List<long> Metadata { get; set; }
```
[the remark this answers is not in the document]
KrzysztofLabus-RDX: You're absolutely right. Renamed that to `metadata_ids`. I guess at some point we should slightly rename some tables and columns having a) short names, b) distinction between "aggregate" and "aggregated", c) unnecessary prefixes in mind.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/EntityStateQuerier.cs
KrzysztofLabus-RDX: Not here.

In general we care about order of things extracted from database where data is paginated and/or should be ordered by "most recently affected". Then obviously we must order it. However we don't really care about order of responses generated for HTTP requests, i.e.: `/entity/details {"addresses":["a", "b"], "explicit_metadata":["x", "y"]` - we don't necessarily have to preserve order of mentioned addresses or explicit metadata.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/EntityStateQuerier.cs
KrzysztofLabus-RDX: I think we're good here. I believe we have a test that validates that?

PawelPawelec-RDX: I'm not sure that we have test that validates order of returned metadata, hmmm can you check with Theo?

PawelPawelec-RDX (approved):

PR #250, opened 2023-04-19
Drop ledger status

KrzysztofLabus-RDX:


src/RadixDlt.NetworkGateway.PostgresIntegration/Migrations/MigrationsDbContextModelSnapshot.cs
```diff
@@ -898,6 +872,22 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                 {
                     b.HasBaseType("RadixDlt.NetworkGateway.PostgresIntegration.Models.Entity");
 
+                    b.Property<string>("BlueprintName")
```
PawelPawelec-RDX, line 875: I'm not sure what happened here and where it came from. I don't see any changes to AccessControllerEntity on that branch, is that some leftover that we forgot to generate migration for?

KrzysztofLabus-RDX: Seems like we must have operate with malformed file previously cause brand-new db migration still results in those changes.

PawelPawelec-RDX (approved):

PR #251, opened 2023-04-20
Added explicit metadata for nested resources

KrzysztofLabus-RDX:


src/RadixDlt.NetworkGateway.PostgresIntegration/Services/EntityStateQuerier.cs
```diff
@@ -1456,4 +1518,24 @@ LIMIT 1
             .Select(e => new { e.Id, e.Address, e.GlobalAddress })
             .ToDictionaryAsync(e => e.Id, e => e.GlobalAddress ?? e.Address.ToHex(), token);
     }
+
+    private async Task<Dictionary<string, long>> ResolveMissingEntityIds(ICollection<GatewayModel.FungibleResourcesCollection>? fungibleResources, ICollection<GatewayModel.NonFungibleResourcesCollection>? nonFungibleResources, CancellationToken token)
```
PawelPawelec-RDX, line 1522: Can we just rename it to ResolveResourceEntityIds? and pass one collection/array of resources regardless if they are fungible/non fungible? it just add complexity and we don't care about it.

KrzysztofLabus-RDX: Renamed to `ResolveResourceEntityIds` but unfortunately we have to pass two collections as their types aren't correlated.

PR #261, opened 2023-04-28
feat: create aggregated GatewayApiClient for SDK

dawidsowardx:


PawelPawelec-RDX (approved):

sdk/typescript/lib/subapis/status.ts
```diff
@@ -0,0 +1,13 @@
+import { StatusApi } from "../generated";
+
+export class Status {
+    constructor(public innerClient: StatusApi) {}
+
+    getGatewayStatus() {
```
dhedey, line 6: Do you think it's worth explicitly annotating types on method returns so that it's clearer if people click through? :)

sdk/typescript/lib/subapis/transaction.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  getStatus(transactionIntentHashHex: string) {
+    return this.innerClient.transactionStatus({
+      transactionStatusRequest: {
+        intent_hash_hex: transactionIntentHashHex,
+      },
+    })
+  }
+
+  getCommitedDetails(
```
dhedey, line 14: Committed*

sdk/typescript/lib/subapis/transaction.ts
```diff
@@ -0,0 +1,31 @@
...
+    transactionIntentHashHex: string,
+    stateVersion?: number
```
dhedey, line 16: Why would you imagine they'd use the state version?

Relatedly - could we add JS Docs to all the methods explaining how they should be used / what they do?

sdk/typescript/README.md
```diff
@@ -1,25 +1,83 @@
...
+
+### Get Transaction Status
+
+```typescript
+const txStatus = await gatewayApi.transaction.getStatus(
+  '266cdfe0a28a761909d04761cdbfe33555ee5fdcf1db37fcf71c9a644b53e60b'
+)
+```
+
+## Fetch polyfill
```
dhedey, line 81: Instead of this section, can we more clearly talk through how to use the library on Node.js - like the docs in the Core API? (eg mentioning the need to use 2.x and passing in the http agents).

Ideally, I think this should be near the top of the doc, in a section about installation / set-up or similar :).

sdk/typescript/README.md
```diff
@@ -1,25 +1,83 @@
...
+  'account_tdx_21_1p823h2sq7nsefkdharvvh5'
+)
+```
+
+## Transaction
+
+### Get Transaction Status
+
+```typescript
+const txStatus = await gatewayApi.transaction.getStatus(
```
dhedey, line 76: Nice :)

Could we show the user doing something with the status? EG console.log'ing the returned status message?

dhedey (commented):
Really nice, thanks.

Just another thought as well - `develop` on the Gateway I don't believe is compatible with `kisharnet` any more - so we probably want to instead fork off and merge into the `release/rcnet-v1` branch, and publish from there - to save confusing people with an SDK which doesn't work against RCnet-v1? :)

sdk/typescript/lib/subapis/transaction.ts
dawidsowardx: Dashboard uses it in order to display ledger state at the moment of submitting TX. We first query by TxID, then get state version, then query gateway again with that state version in order to get `epoch` and `round` at the time TX was submitted.

Now when you're asking I'm thinking I maybe should wrap those two calls inside SDK?

sdk/typescript/lib/subapis/status.ts
dawidsowardx: Hm... Yes I think it is

dawidsowardx:
I'm additionally wondering if we shouldn't by default convert all object keys to camelCase instead of snake_case @dhedey

sdk/typescript/lib/subapis/transaction.ts
dhedey: It sounds like either we should change committed details to return those details, or yes, we create some helper which does these two queries for you  - or maybe separate the queries?

I'm surprised we don't return the ledger state identifiers (eg epoch/round/state version) of a committed transaction already? Maybe this would be an easy request to the gateway team if it's not there.

dawidsowardx: Gateway does return **current** ledger state whereas what we want is ledger state at the moment of tx submission. I guess I could use another endpoint that basically returns ledger state (epoch&round) for given state version.

sdk/typescript/README.md
dawidsowardx: I didn't have to include any special flag, I think it's turned on by default. All i did is saving this into `index.js` and then doing `node index.js`

dawidsowardx:
> > I'm additionally wondering if we shouldn't by default convert all object keys to camelCase instead of snake_case @dhedey
> 
> I think it's a lot of work, possibly without much gain - as it means the objects no longer align with the API docs 🤷. But up for being convinced otherwise. Maybe it's a simple flag on the client generation.

I imagine it as one recursive function that  converts everything automatically. We should also be able to write Typescript generic that converts one into another recursively

PR #275, opened 2023-05-05
DO-1241 / main / replace gh secrets with aws

kofrdx:


sdk/typescript/lib/subapis/state.ts
```diff
@@ -0,0 +1,65 @@
+import {
```
shambupujar, line 1: There are code changes here. Not sure if this is intended in this branch?

PR #280, opened 2023-05-08
Feature/add total supply min burn to resources

PawelPawelec-RDX:


src/RadixDlt.CoreApiSdk/Model/EventTypeIdentifiers.cs
```diff
@@ -0,0 +1,86 @@
...
+ * You are solely responsible for; testing, validating and evaluation of all operation
+ * logic, functionality, security and appropriateness of using the Work for any commercial
+ * or non-commercial purpose and for any reproduction or redistribution by You of the
+ * Work. You assume all risks associated with Your use of the Work and the exercise of
+ * permissions under this License.
+ */
+
+namespace RadixDlt.CoreApiSdk.Model;
+
+public static class EventTypeIdentifiers
```
KrzysztofLabus-RDX, line 67: This is problematic a bit cause those are not constants, those are not well-known values. We should most likely obtain them during genesis TX processing and store them in `network_configuration` table or something...

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/EntityStateQuerier.cs
```diff
@@ -170,15 +172,23 @@ public async Task<GatewayModel.StateEntityDetailsResponse> EntityDetails(List<Gl
             switch (entity)
             {
                 case FungibleResourceEntity frme:
+                    var fungibleResourceSupplyData = resourcesSupplyData!.Single(x => x.ResourceEntityId == frme.Id);
```
KrzysztofLabus-RDX, line 175: null-negation stinks... guess we might want to assign "empty dictionary" rather null then?

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/EntityStateQuerier.cs
```diff
@@ -800,6 +810,37 @@ LIMIT 1
...
+) resh ON true
+",
+            parameters: new
+            {
+                stateVersion = ledgerState.StateVersion,
+                entityIds = entityIds,
+            },
+            cancellationToken: token);
+
+        var result = (await _dbContext.Database.GetDbConnection().QueryAsync<ResourceSupplyViewModel>(cd)).ToList();
```
KrzysztofLabus-RDX, line 840: Why don't we just use EF Core and `.FromSqlInterpolated`? It looks like we could use already existing entity type definitions? Guess this custom `ResourceSupplyViewModel` is needless?

src/RadixDlt.CoreApiSdk/Model/EventTypeIdentifiers.cs
KrzysztofLabus-RDX: But what I've described above is going to be bit time consuming and is going to require some extra feature in RET we don't have today. IMO it should be moved out of CoreApi contracts (cause it does not belong there) and maybe annotated with TODO?

KrzysztofLabus-RDX (changes requested):

src/RadixDlt.CoreApiSdk/Model/EventTypeIdentifiers.cs
PawelPawelec-RDX: I can move it somewhere but it's CoreAPI contract, I don't see any better place for now where to keep it.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/EntityStateQuerier.cs
```diff
@@ -800,6 +821,23 @@ LIMIT 1
...
+FROM variables
+INNER JOIN LATERAL(
+    SELECT *
+    FROM resource_entity_supply_history
+    WHERE from_state_version <= {ledgerState.StateVersion} AND resource_entity_id = variables.entity_id
+    ORDER BY from_state_version DESC
+    LIMIT 1
+) resh ON true").AsNoTracking().ToListAsync(token);
+
+        return result;
```
KrzysztofLabus-RDX, line 838: Isn't `_dbContext` a `ReadOnlyDbContext`? It's `.AsNoTracking()` by default so it's not needed to specify that explicitly.

KrzysztofLabus-RDX (changes requested):

PR #307, opened 2023-06-02
Release/ash

KrzysztofLabus-RDX:


PawelPawelec-RDX (approved):

Theoklitos:
The e2e test failure here is expected. 

Unfortunately, it tries to use the `release/ash` client to run the tests, which expects a different contract, so the tests will fail. I don't think our workflow can handle merging back into `develop` from release branches.

PR #322, opened 2023-06-20
Tuned metadata parsing for Birch

KrzysztofLabus-RDX:


dhedey (approved):
Approving - but can I emphasis that we really need to update the metadata models in the Gateway to be structural / include the type information - because projects and Core Apps teams keep requesting that they need to see the type information for their own uses to handle them differently.

So ideally we'd just have models which parallel the metadata data model itself

PR #332, opened 2023-06-27
fix: error IL3000

kofrdx:
I have checked through the babylon-gateway release workflow and could not get the build to work. Now I am very bad when it comes to dotnet. Therefor I was hoping for your opinion on this error and fix.

This command:
```
dotnet publish --runtime linux-x64 --self-contained false -p:PublishSingleFile=true -p:PublishReadyToRun=true -p:DebugType=None -p:DebugSymbols=false "apps/DataAggregator/DataAggregator.csproj" -c Release -o /tmp/app/dataaggregator
```
Fails with this error message:
```
/Users/kim.fehrs/projects/rdx/radixdlt-org/babylon-gateway/src/RadixDlt.NetworkGateway.PostgresIntegration/Services/LedgerStateQuerier.cs(244,61): error IL3000: 'System.Reflection.Assembly.Location' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'. [/Users/kim.fehrs/projects/rdx/radixdlt-org/babylon-gateway/src/RadixDlt.NetworkGateway.PostgresIntegration/RadixDlt.NetworkGateway.PostgresIntegration.csproj]
```
Which is found [here](https://github.com/radixdlt/babylon-gateway/blob/91fba72f87567a052302955f91cfcc045ca9bb44/src/RadixDlt.NetworkGateway.PostgresIntegration/Services/LedgerStateQuerier.cs#L244).

The search result for the error gave me [this](https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/warnings/il3000)

And the line would look like this after the proposed fix:

```
var productVersion = FileVersionInfo.GetVersionInfo(AppContext.BaseDirectory).ProductVersion; 
```

Afterward it works for single-file builds with

```
dotnet publish --runtime linux-x64 --self-contained false -p:PublishSingleFile=true -p:PublishReadyToRun=true -p:DebugType=None -p:DebugSymbols=false "apps/DataAggregator/DataAggregator.csproj" -c Release -o /tmp/app/dataaggregator
```

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -20,11 +20,11 @@ jobs:
       - name: Build linux binaries
         run: |
           cd apps/DataAggregator
-          dotnet publish --runtime linux-x64 --configuration Release --self-contained false -p:PublishSingleFile=true -p:PublishReadyToRun=true -p:DebugType=None -p:DebugSymbols=false --output ./output
+          dotnet publish --runtime linux-x64 --configuration Release --self-contained false -p:PublishReadyToRun=true -p:DebugType=None -p:DebugSymbols=false --output ./output
           cd ../DatabaseMigrations
-          dotnet publish --runtime linux-x64 --configuration Release --self-contained false -p:PublishSingleFile=true -p:PublishReadyToRun=true -p:DebugType=None -p:DebugSymbols=false --output ./output
+          dotnet publish --runtime linux-x64 --configuration Release --self-contained false -p:PublishReadyToRun=true -p:DebugType=None -p:DebugSymbols=false --output ./output
           cd ../GatewayApi
-          dotnet publish --runtime linux-x64 --configuration Release --self-contained false -p:PublishSingleFile=true -p:PublishReadyToRun=true -p:DebugType=None -p:DebugSymbols=false --output ./output
+          dotnet publish --runtime linux-x64 --configuration Release --self-contained false -p:PublishReadyToRun=true -p:DebugType=None -p:DebugSymbols=false --output ./output
           cd ../..
           zip -r data-aggregator.zip apps/DataAggregator/output/
           zip -r database-migrations.zip apps/DatabaseMigrations/output/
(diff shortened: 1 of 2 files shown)


KrzysztofLabus-RDX:
Actually I think we should ditch the idea of `-p:PublishSingleFile=true` as there's no good reason or benefits to actually use it. What's more it could lead to some undesired behaviors as one you've found.

kofrdx:
@KrzysztofLabus-RDX 
Who do we need for that decission?  Because that would affect gateway operators in the way they do installations/updates right?

KrzysztofLabus-RDX:
@kofrdx It does not really changes much and it should not even have any impact on further "run/installation" commands. It should be perfectly fine to just remove this argument. What it does is that it combines all DLL files into one, i.e.:

```
-rwxr--r--  1 krzysztof.labus  staff    67960 Sep 13  2019 Microsoft.Extensions.DependencyModel.dll
-rwxr--r--  1 krzysztof.labus  staff   169096 Aug 20  2022 Microsoft.Extensions.Identity.Core.dll
-rwxr--r--  1 krzysztof.labus  staff    42120 Aug 20  2022 Microsoft.Extensions.Identity.Stores.dll
-rwxr--r--  1 krzysztof.labus  staff    62824 Mar 30  2021 Microsoft.IdentityModel.JsonWebTokens.dll
-rwxr--r--  1 krzysztof.labus  staff    24936 Mar 30  2021 Microsoft.IdentityModel.Logging.dll
-rwxr--r--  1 krzysztof.labus  staff   108920 Mar 30  2021 Microsoft.IdentityModel.Protocols.OpenIdConnect.dll
-rwxr--r--  1 krzysztof.labus  staff    32632 Mar 30  2021 Microsoft.IdentityModel.Protocols.dll
-rwxr--r--  1 krzysztof.labus  staff   889216 Mar 30  2021 Microsoft.IdentityModel.Tokens.dll
-rwxr--r--  1 krzysztof.labus  staff   531968 Sep 16  2022 Npgsql.EntityFrameworkCore.PostgreSQL.dll
....
-rwxr--r--  1 krzysztof.labus  staff   531968 Sep 16  2022 DataAggreggator
```

gets reduced to a single
```
-rwxr--r--  1 krzysztof.labus  staff   531968 Sep 16  2022 DataAggreggator
```

which is like a self-extracting ZIP file containing all of those DLLs and binaries

kofrdx:
@KrzysztofLabus-RDX 
I updated the PR to that solution. Sounds like the most stable way.

KrzysztofLabus-RDX (approved):

PR #336, opened 2023-07-03
added validator uptime.

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -2033,6 +2045,16 @@ components:
               description: This cursor allows forward pagination, by providing the cursor from the previous request.
               type: string
               nullable: true
+            uptime_range:
```
KrzysztofLabus-RDX, line 2048: /How about `D90`, `D30`, ..., `H12`? Identifiers (enum values) cannot start with digit on any sane platform so this would result in cleaner code.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
```diff
@@ -947,13 +965,47 @@ private async Task<ExtendLedgerReport> ProcessTransactions(ReadWriteDbContext db
...
                             continue;
                         }
 
                         var eventEmitterEntity = referencedEntities.Get((EntityAddress)methodEventEmitter.Entity.EntityAddress);
 
+                        if (methodEventEmitter.Entity.EntityType == CoreModel.EntityType.GlobalValidator)
+                        {
+                            if (@event.Type.LocalTypeIndex.Index == eventTypeIdentifiers.GlobalValidator.ValidatorEmissionApplied)
+                            {
+                                var data = (JObject)@event.Data.ProgrammaticJson;
```
KrzysztofLabus-RDX, line 980: BTW maybe we should decompile SBOR using RET.NET into something easier to ingest? Guess we might even need to define those event schemas... it's impossible to rely on `data[fields][4]` for anything long term. 

Ah... actually those events and their schema should already be available in package definition but until we get this "programmatic json with property names" we're doomed. Let's keep it this way... maybe just annotate it with TODO?

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/WriteHelper.cs
```diff
@@ -440,6 +440,33 @@ public async Task<int> CopyValidatorActiveSetHistory(ICollection<ValidatorActive
...
+
+        foreach (var e in entries)
+        {
+            await writer.StartRowAsync(token);
+            await writer.WriteAsync(e.Id, NpgsqlDbType.Bigint, token);
+            await writer.WriteAsync(e.ValidatorEntityId, NpgsqlDbType.Bigint, token);
+            await writer.WriteAsync(e.FromStateVersion, NpgsqlDbType.Bigint, token);
+            await writer.WriteAsync(e.EpochNumber, NpgsqlDbType.Bigint, token);
+            await writer.WriteAsync(e.EpochStart, NpgsqlDbType.TimestampTz, token);
+            await writer.WriteAsync(e.EpochEnd, NpgsqlDbType.TimestampTz, token);
```
KrzysztofLabus-RDX, line 460: Open question: maybe we should introduce new table `epoch` with `epoch_number, epoch_start_date, epoch_end_date, validators, ...`?

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/EntityStateQuerier.cs
```diff
@@ -612,8 +626,27 @@ LIMIT 1
 
         var cd = new CommandDefinition(
             commandText: @"
-WITH variables (validator_entity_id) AS (SELECT UNNEST(@validatorIds))
-SELECT e.id AS ValidatorId, CAST(evh.balance AS text) AS Balance, esh.state AS State, evh.from_state_version AS BalanceLastUpdatedAtStateVersion, esh.from_state_version AS StateLastUpdatedAtStateVersion
+WITH
+variables (validator_entity_id) AS (SELECT UNNEST(@validatorIds)),
+validator_uptime_aggregated (validator_entity_id, proposals_made, proposals_missed, epochs_active_in) AS (
```
KrzysztofLabus-RDX, line 631: Open question: use `SUM(...) AS proposals_made_sum` to improve readability?

KrzysztofLabus-RDX (changes requested):

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/WriteHelper.cs
PawelPawelec-RDX: hmm we could but i'm not sure if it is not better to have denormalized read model here. Are we using that in any other place?

PR #344, opened 2023-07-13
Use unambiguous metadata value type definitions in OAS

KrzysztofLabus-RDX:


dhedey (approved):
Nice, very carefully done. Although looks like something is broken in `tests/RadixDlt.NetworkGateway.UnitTests/PostgresIntegration/ScryptoSborUtilsTests.cs`

PR #345, opened 2023-07-13
Merge birch updates

dawidsowardx:


sdk/typescript/lib/constants.ts
```diff
@@ -2,3 +2,8 @@
  * Maximum number of addresses that can be queried at once when using /state/entity/details endpoint
  */
 export const MAX_ADDRESSES_COUNT = 20
+
+/**
+ * Maximum number of NFT IDs that can be queried at once when using /state/non-fungible/data endpoint
+ */
+export const MAX_NFT_IDS_COUNT = 29
```
KrzysztofLabus-RDX, line 9: Those should not be constants, they could be defined as default values at most. NG is configurable when it comes to these.

KrzysztofLabus-RDX (approved):

PR #347, opened 2023-07-14
validators uptime

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -555,6 +555,28 @@ paths:
         "4XX":
           $ref: "#/components/responses/ErrorResponse"
 
+  /state/validators/uptime:
```
KrzysztofLabus-RDX, line 558: I think validator update should not be part of `state/*` endpoint group as it has nothing to do with the ledger state? @dhedey ?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1133,6 +1155,37 @@ components:
         - Native
         - ScryptoV1
 
+    ValidatorUptimeCollection:
```
KrzysztofLabus-RDX, line 1158: I'd say validator update endpoint should accept a collection of validator addresses we're interested in. No need for pagination.

src/RadixDlt.NetworkGateway.PostgresIntegration/CommonDbContext.cs
```diff
@@ -131,7 +131,7 @@ internal abstract class CommonDbContext : DbContext
 
     public DbSet<PackageBlueprintHistory> PackageBlueprintHistory => Set<PackageBlueprintHistory>();
 
-    public DbSet<ComponentSchema> ComponentSchema => Set<ComponentSchema>();
+    public DbSet<ValidatorUptime> ValidatorsUptime => Set<ValidatorUptime>();
```
KrzysztofLabus-RDX, line 134: Naming: this table does not store validator uptime; it stores some statistics generated at the end of the epoch; we compute uptime based on those "generic" stats, am I right? 

Sounds like some "validator epoch statistics history" to me?

KrzysztofLabus-RDX (changes requested):

src/RadixDlt.NetworkGateway.PostgresIntegration/CommonDbContext.cs
PawelPawelec-RDX: hmm ok, I think we better rename it to ValidatorEmissions, that's how engine calls that.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/EntityStateQuerier.cs
```diff
@@ -755,6 +757,66 @@ LIMIT 1
         return new GatewayModel.StateValidatorsListResponse(ledgerState, new GatewayModel.ValidatorCollection(null, null, nextCursor, items));
     }
 
+    public async Task<GatewayModel.StateValidatorsUptimeResponse> StateValidatorsUptime(GatewayModel.StateValidatorsUptimeCursor? cursor, GatewayModel.LedgerState ledgerState,
+        GatewayModel.LedgerState? fromLedgerState, CancellationToken token = default)
+    {
+        var validatorsPageSize = _endpointConfiguration.Value.ValidatorsPageSize;
```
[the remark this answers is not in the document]
PawelPawelec-RDX: hmm we do exactly same approach in almost all other methods here. To be honest i'm not sure what responsibility EntityStateQuerier has and hasn't.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: hmm are you sure about that requirement?

PawelPawelec-RDX: that would mean, we need to get these addresses first, also passing 100 or more addresses doesn't sound reasonable

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
dhedey: So long as it takes a state version and is deterministic against that state version, I'm happy with it being in state.

But also happy with it being in a different place, eg `/aggregate/` or `/statistics/` or `/data/` or something.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
dhedey: The requirement is that it is easy (read - not too many queries) to get validator information and uptime for all validators (including non-active validators).

If there's a strong desire not to have it on the validator models, then we should at least make sure we can clearly explain how the validators page on the dashboard is intended to retrieve this information.

Accepting a (possibly large-ish) collection of validator addresses and letting the front end join the two queries up is reasonable, as it would also allow this endpoint to be used for a single or small number of validators.

KrzysztofLabus-RDX (approved):

PR #353, opened 2023-07-19
Include module in access rules

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
```diff
@@ -1393,13 +1393,14 @@ private async Task<ExtendLedgerReport> ProcessTransactions(ReadWriteDbContext db
 
                 foreach (var entry in accessRuleChange.Entries)
                 {
-                    var entryLookup = new AccessRuleEntryLookup(lookup.EntityId, entry.Key.RoleKey);
+                    var entryLookup = new AccessRuleEntryLookup(lookup.EntityId, entry.Key.RoleKey, entry.Key.ObjectModuleId.ToString());
```
KrzysztofLabus-RDX, line 1396: `ObjectModuleId` is an enum and we should map it as such in our database.

PawelPawelec-RDX: addressed

PR #390, opened 2023-08-04
return effective role assignments for native modules and resource ent…

PawelPawelec-RDX:
…ities.


src/RadixDlt.NetworkGateway.PostgresIntegration/Services/RoleAssignmentsMapper.cs
```diff
@@ -0,0 +1,151 @@
...
+
+    public RoleAssignmentsMapper(IRoleAssignmentsKeyProvider roleAssignmentsKeyProvider)
+    {
+        _roleAssignmentsKeyProvider = roleAssignmentsKeyProvider;
+    }
+
+    public Dictionary<long, GatewayModel.ComponentEntityRoleAssignments> GetEffectiveRoleAssignments(
+        ICollection<Entity> componentEntities, ICollection<EntityRoleAssignmentsOwnerRoleHistory> ownerRoles, ICollection<EntityRoleAssignmentsEntryHistory> roleAssignments)
+    {
+        var fungibleResourceAccessRuleKeys = _roleAssignmentsKeyProvider.GetFungibleResourceKeys();
```
dhedey, line 95: Minor, but I'd suggest moving the fungible / nonFungible resource key creation into the relevant if block, for a slight performance improvement to avoid some allocations in the 90% case where it's not needed.

dhedey (approved):
Looks great! A few minor thoughts, all optional.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/RoleAssignmentsMapper.cs
PawelPawelec-RDX: I'm not sure about that, that method maps multiple entities, so we'll call that method potentially 30x times if all entities will be fungible resources. It's not a big deal for now since it's all stored in memory and read once, will need rework anyway once we start reading keys from blueprint definition.

PR #423, opened 2023-08-28
Update the Dockerfiles to run as non-root

ghost:


shambupujar (approved):

krzlabrdx:
@muzuke @shambupujar Wouldn't it be better to name group/user say `app` or similar? Guess it does not actually has to be that precise/specific.

marek-karwacki-rdx (approved):

PR #479, opened 2023-09-14
Fix event schema lookup

krzlabrdx:


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+### 0.5.4 RCNet v3.1 revision 3
+Release Date: 14.09.2023
+
+- Fix event schema lookup in `/stream/transactions`
```
PawelPawelec-RDX, line 4: transaction/committeddetails as well.

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/LedgerTransaction.cs
```diff
@@ -179,6 +184,48 @@ internal class TransactionReceipt
...
+
+        for (var i = 0; i < EventSbors.Length; ++i)
+        {
+            result.Add(new TransactionReceiptEventLookup(EventSchemaEntityIds[i], EventSchemaHashes[i]));
+        }
+
+        return result;
+    }
+
+    public List<TransactionReceiptEventData> GetEvents()
```
PawelPawelec-RDX, line 206: we already had similar extension TransactionReceiptExtensions with type TransactionReceiptEventData which is probably unused now.

CHANGELOG.md
krzlabrdx: updated

PawelPawelec-RDX (approved):

PR #484, opened 2023-09-15
feature: Pending transactions rework

dhedey:
Testing performed:
* Submit invalid transaction
* Submit valid transaction. Check DB and status.
* Submit duplicate valid transaction. Check DB and status.
* Submit transaction with down Core API on Gateway, check interim API result and check gets resubmitted by Aggregator. Use DB to check that resubmission sees exponential back-off.
* Adjusted pruning delay and checked it works


src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
```diff
@@ -0,0 +1,240 @@
...
+
+            return result.FailureResponse.Details switch
+            {
+                TransactionSubmitIntentAlreadyCommitted alreadyCommittedDetails =>
+                    new NodeSubmissionResult.IntentAlreadyCommitted(alreadyCommittedDetails.CommittedAs),
+                TransactionSubmitPriorityThresholdNotMetErrorDetails priorityThresholdDetails =>
+                    new NodeSubmissionResult.MempoolPriorityThresholdNotMet(priorityThresholdDetails),
+                TransactionSubmitRejectedErrorDetails rejectedDetails =>
+                    new NodeSubmissionResult.Rejection(rejectedDetails),
+                _ => throw new ArgumentOutOfRangeException(nameof(result.FailureResponse.Details)),
```
krzlabrdx, line 129: Please use at least `new ArgumentOutOfRangeException(nameof(...), details, null)` - otherwise we won't know which unexpected/unhandled type was actually observed.

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
```diff
@@ -0,0 +1,240 @@
...
+            };
+        }
+        catch (OperationCanceledException ex)
```
krzlabrdx, line 132: We need to make this slightly more robust. We should catch and test if it was our CT that triggered the exception: `catch (OperationCancelledEx ex) when (cts.Token.IsCancellationRequested)` otherwise we might handle non-timeout related exceptions.

EDIT:
```
catch (OpertionCancelledEx ex) when (cts.token.IsCancellatin)
  // return "operation cancelled by internal timeout of 123 seconds"
catch (OperationCancelledEx ex)
  // return "operation cacnelled by something else"
```

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
```diff
@@ -0,0 +1,240 @@
...
+        {
+            return new NodeSubmissionResult.OtherSubmissionError(ex, "Unknown error");
+        }
+    }
+}
+
+public abstract record NodeSubmissionResult
+{
+    // ======================================================================
+    // WARNING WHEN ADDING VARIANTS:
```
krzlabrdx, line 146: While Rust and its pattern matching is no doubt a great technique, C# isn't designed this way. IMO we should avoid those constructs if possible. Otherwise this codebase is going to become super error-prone maintenance nightmare.

For instance `ShouldStopSubmittingPermanently` should be simply overridden by implementations to return their specific logic; we don't need type-based pattern matching here. Same's true for intent and payload status.

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
```diff
@@ -0,0 +1,240 @@
...
+    public string? GetExecutionFailureReason()
+    {
+        return this switch
+        {
+            Rejection rejection => rejection.Details.ErrorMessage,
+            _ => null,
+        };
+    }
+
+    public string? GetApiSubmitErrorMessage()
```
krzlabrdx, line 222: IMO it would be better to throw exception here if we're dealing with successful result. Client to this piece of code should not call this method on success.

i.e. avoid nulls if possible.

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
```diff
@@ -0,0 +1,240 @@
...
+}
+
+public interface ITransactionSubmitterObserver
+{
+    ValueTask ObserveSubmitAttempt(SubmitContext context);
+
+    ValueTask ObserveSubmitResult(SubmitContext context, NodeSubmissionResult nodeSubmissionResult);
+}
+
+public record SubmitContext(bool IsResubmission);
```
krzlabrdx, line 240: Hmmm... shouldn't this context type also convey notarized TX blob and network id/name?

src/RadixDlt.NetworkGateway.Abstractions/Model/PendingTransactionIntentLedgerStatus.cs
```diff
@@ -0,0 +1,136 @@
...
+    /// </summary>
+    CommittedSuccess,
+
+    /// <summary>
+    /// We know that the intent has been committed as a failure.
+    /// </summary>
+    CommittedFailure,
+
+    /// <summary>
+    /// We know that the intent has been committed, but don't yet know the outcome until it's synced (Success/Failure).
```
krzlabrdx, line 85: Open question: how can we tell something has been committed not seeing it in the committed TX stream? My guess is that this is going to be driven by the fact that we've seen given TX in mempool and then it disappeared? This does not sound like enough information to conclude that the TX got committed to me.

src/RadixDlt.NetworkGateway.Abstractions/Model/PendingTransactionIntentLedgerStatus.cs
```diff
@@ -0,0 +1,136 @@
...
+    ///
+    /// <para>Or the payload could have been submitted with the wrong signatures, and the notary corrects
+    /// these and resubmits the intent in a new payload, which can be submitted.</para>
+    /// </summary>
+    LikelyButNotCertainRejection,
+
+    /// <summary>
+    /// We don't know anything about whether the intent could be committed or rejected.
+    /// </summary>
+    Unknown,
```
krzlabrdx, line 114: Please make it the very first value so that it is translated to numerical value of `0`.

src/RadixDlt.NetworkGateway.Abstractions/Model/PendingTransactionIntentLedgerStatus.cs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+public static class PendingTransactionIntentLedgerStatusExtensions
+{
+    /// <summary>
+    /// A new status should overwrite an older status only if it has >= ReplacementPriority.
+    /// </summary>
+    public static int ReplacementPriority(this PendingTransactionIntentLedgerStatus status)
```
krzlabrdx, line 122: Holy moly! This function strongly indicates that we must stop trying to override/update any information stored in database and that we need eventsourcing-like database and asychronous state machine for pending TX area. Not that we need to drop RDBMS, just to change the ways of working with this concept.

src/RadixDlt.NetworkGateway.Abstractions/Model/PendingTransactionPayloadLedgerStatus.cs
```diff
@@ -0,0 +1,162 @@
...
+
+    /// <summary>
+    /// This specific payload was rejected at the last execution.
+    /// </summary>
+    TransientlyRejected,
+
+    /// <summary>
+    /// We don't know anything about whether the intent could be committed or rejected.
+    /// </summary>
+    Unknown,
```
krzlabrdx, line 117: Similarly to intent counterpart, please move it as a very first element of the enum.

src/RadixDlt.NetworkGateway.Abstractions/Model/PendingTransactionPayloadLedgerStatus.cs
```diff
@@ -0,0 +1,162 @@
...
+            PendingTransactionPayloadLedgerStatus.CommittedSuccess => true,
+            PendingTransactionPayloadLedgerStatus.CommittedFailure => true,
+            PendingTransactionPayloadLedgerStatus.CommitPendingOutcomeUnknown => true,
+            PendingTransactionPayloadLedgerStatus.CommitOfOtherPayloadForIntentPendingOutcomeUnknown => true,
+            PendingTransactionPayloadLedgerStatus.PermanentlyRejected => true,
+            PendingTransactionPayloadLedgerStatus.TransientlyAccepted => false,
+            PendingTransactionPayloadLedgerStatus.TransientlyRejected => false,
+            PendingTransactionPayloadLedgerStatus.Unknown => false,
+            _ => throw new ArgumentOutOfRangeException(nameof(status), status, null),
+        };
```
krzlabrdx, line 135: That's a very Rusty way of coding :D

src/RadixDlt.NetworkGateway.DataAggregator/Services/LedgerConfirmationService.cs
```diff
@@ -457,6 +460,7 @@ private void UpdateRecordsOfTopOfLedger(TransactionSummary topOfLedger)
         _observers.ForEach(x => x.RecordTopOfDbLedger(topOfLedger.StateVersion, topOfLedger.RoundTimestamp));
 
         _systemStatusService.SetTopOfDbLedgerNormalizedRoundTimestamp(topOfLedger.NormalizedRoundTimestamp);
+        _transactionService.SetCurrentEpoch(topOfLedger.Epoch);
```
krzlabrdx, line 463: Please don't... this code is becoming more and more entangled. Ledger confirmation service instructing completely unrelated pending TX service about current epoch? This one should be decoupled.

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
dhedey: 👍 true - better to make these abstract methods

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
dhedey: Possibly 👍

src/RadixDlt.NetworkGateway.Abstractions/Model/PendingTransactionIntentLedgerStatus.cs
dhedey: No. This is because a Core API has told us that it's been committed at submission time, we just haven't synced it yet.

src/RadixDlt.NetworkGateway.Abstractions/Model/PendingTransactionIntentLedgerStatus.cs
dhedey: 👍

src/RadixDlt.NetworkGateway.DataAggregator/Services/LedgerConfirmationService.cs
dhedey: Yeah this was a bit of a 4am "make it work". Would you be up for me registering the PendingTransactionService as a `LedgerStateTipObserver`?

src/RadixDlt.NetworkGateway.Abstractions/Model/PendingTransactionPayloadLedgerStatus.cs
dhedey: Hah yes! Will change to an abstract method.

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
dhedey: Maybe? Depends how this is used. Will review.

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
dhedey: That said, I am a big fan of allowing pattern matching, and capturing all possible outcomes into such a sum type. It provides better encapsulation of the concept of a result, and you're less likely to miss things. (I accept that we should avoid switches ofc, your suggestion to use abstract methods is better).

EG see how nicely abstracted the metrics are now, and it makes it much easier to handle all cases, instead of missing eg an Exception.

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
dhedey: Nice thoughts 👍

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
```diff
@@ -0,0 +1,240 @@
...
+ *
+ * You are solely responsible for; testing, validating and evaluation of all operation
+ * logic, functionality, security and appropriateness of using the Work for any commercial
+ * or non-commercial purpose and for any reproduction or redistribution by You of the
+ * Work. You assume all risks associated with Your use of the Work and the exercise of
+ * permissions under this License.
+ */
+
+using RadixDlt.CoreApiSdk.Api;
+using RadixDlt.CoreApiSdk.Model;
```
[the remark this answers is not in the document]
dhedey: 👍

src/RadixDlt.NetworkGateway.Abstractions/Model/PendingTransactionPayloadLedgerStatus.cs
dhedey: Actually I can't - because it's an enum. But will change the `editorconfig` to allow us to make enum matching of named variants exhaustive, so a change will be caught as a compiler error.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
```diff
@@ -182,26 +182,33 @@ private async Task<int> UpdatePendingTransactions(ReadWriteDbContext dbContext,
          *
          * and then simple foreach loop notifying the observers
          */
-
-        var userTransactionStatusByPayloadHash = new Dictionary<string, PendingTransactionStatus>(committedTransactions.Count);
+        var timestamp = _clock.UtcNow;
+        var userTransactionStateVersionAndSuccessByPayloadHash = new Dictionary<string, (long, bool, string?)>(committedTransactions.Count);
```
krzlabrdx, line 186: Please define some record class for this `(long, bool, string?)` tuple.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/PendingTransactionPrunerService.cs
```diff
@@ -82,94 +83,63 @@ internal class PendingTransactionPrunerService : IPendingTransactionPrunerServic
...
         await using var dbContext = await _dbContextFactory.CreateDbContextAsync(token);
 
-        var rawCountByStatus = await dbContext
+        var countByStatus = await dbContext
             .PendingTransactions
-            .GroupBy(t => t.Status)
-            .Select(g => new { Status = g.Key, Count = g.Count() })
+            .GroupBy(canPruneExpression)
+            .Select(g => new PendingTransactionCountByPruneStatus(g.Key, g.Count()))
             .ToListAsync(token);
```
krzlabrdx, line 120: IIRC EF Core struggled a bit when we used custom type (`PendingTransactionCountByPruneStatus`) in select expression. Hence the `rawCountByStatus` followed by another LINQ expression. Could you confirm this refactored code does not result in warnings?

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/TransactionQuerier.cs
```diff
@@ -332,14 +332,218 @@ public async Task<TransactionPageWithoutTotal> GetTransactionStream(TransactionS
...
+        }
+    }
+
+    public async Task<GatewayModel.TransactionStatusResponse> ResolveTransactionStatusResponse(GatewayModel.LedgerState ledgerState, string intentHash, CancellationToken token = default)
+    {
+        var maybeCommittedTransactionSummary = await _dbContext
+            .LedgerTransactions
+            .OfType<UserLedgerTransaction>()
+            .Where(ult => ult.StateVersion <= ledgerState.StateVersion && ult.IntentHash == intentHash)
+            .Select(ult => new CommittedTransactionSummary(
```
krzlabrdx, line 510: Again, could you please confirm this no longer produces any warning?

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
```diff
@@ -0,0 +1,274 @@
...
+                CoreModel.TransactionSubmitPriorityThresholdNotMetErrorDetails priorityThresholdDetails =>
+                    new NodeSubmissionResult.MempoolPriorityThresholdNotMet(priorityThresholdDetails),
+                CoreModel.TransactionSubmitRejectedErrorDetails rejectedDetails =>
+                    new NodeSubmissionResult.Rejection(rejectedDetails),
+                _ => throw new ArgumentOutOfRangeException(nameof(result.FailureResponse.Details), result.FailureResponse.Details, null),
+            };
+        }
+        catch (OperationCanceledException ex) when (timeoutTokenSource.IsCancellationRequested)
+        {
+            return new NodeSubmissionResult.OtherSubmissionError(ex, IsTimeout: true, $"Operation timed-out after {submitContext.SubmissionTimeout.TotalSeconds} seconds");
```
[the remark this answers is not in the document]
dhedey: You prefer milliseconds to seconds?  Happy to make the change, although not sure I understand why it's easier to parse, so I might have missed a subtlety?

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
```diff
@@ -0,0 +1,274 @@
...
+        }
+        catch (OperationCanceledException ex)
+        {
+            return new NodeSubmissionResult.OtherSubmissionError(ex, IsTimeout: false, "Operation cancelled (not due to timeout)");
```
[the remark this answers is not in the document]
dhedey: Good point 👍

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
```diff
@@ -0,0 +1,274 @@
...
+    {
+        public override bool ShouldStopSubmittingPermanently() => true;
+
+        public override PendingTransactionPayloadLedgerStatus PayloadStatus() => CommittedAsDetails.IsSameTransaction
+            ? PendingTransactionPayloadLedgerStatus.CommitPendingOutcomeUnknown
+            : PendingTransactionPayloadLedgerStatus.CommitOfOtherPayloadForIntentPendingOutcomeUnknown;
+
+        public override PendingTransactionIntentLedgerStatus IntentStatus() => PendingTransactionIntentLedgerStatus.CommitPendingOutcomeUnknown;
+
+        public override string MetricLabel() => "already_committed";
```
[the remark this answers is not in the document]
dhedey: Right. I had the same label across two switch statements, and was just like. You know what, let's be pragmatic about this.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/PendingTransactionPrunerService.cs
dhedey: Seems to be okay - can't see any logs... What warnings specifically were you getting?

src/RadixDlt.NetworkGateway.Abstractions/CoreCommunications/TransactionSubmitter.cs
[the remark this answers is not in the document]
krzlabrdx: Ouh, this is not a log entry message but user-facing instruction. OK, forget it... guess seconds are just fine. ;-)

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/PendingTransactionPrunerService.cs
krzlabrdx: Honestly speaking don't remember what was the exact issue. It complaint about this custom type in `.Select` - this is why we used this two-step procedure to first fetch the data into anonymous object and then to map it into desired custom type. Maybe with recent EF Core upgrades they solve this inconvenience? If it works just fine I'm more than happy to have it coded in this simplified fashion! :)

dhedey: Seems to work fine :D

krzlabrdx (approved):

krzlabrdx (approved):

PR #495, opened 2023-09-22
Performance-related metrics / resolved TODOs

krzlabrdx:


dhedey (approved):
Approved. Although we might want to configure our histogram buckets

PR #496, opened 2023-09-22
Upgrade to babylon

krzlabrdx:


src/RadixDlt.NetworkGateway.DataAggregator/Workers/NodeWorkers/NodeTransactionFetchWorker.cs
```diff
@@ -236,10 +238,31 @@ private async Task<FetchedTransactions> FetchTransactions(long fromStateVersion,
...
+        });
 
-        await _observers.ForEachAsync(x => x.TransactionsFetched(NodeName, transactions, fetchTransactionsMs));
+        if (apiResponse.Failed)
+        {
+            var ex = apiResponse.FailureResponse.OriginalApiException;
+
+            if (apiResponse.FailureResponse.Details is CoreModel.RequestedStateVersionOutOfBoundsErrorDetails oob)
+            {
+                _logger.LogWarning(ex, "Requested state version past the maximal ledger state version of {MaxLedgerStateVersion}", oob.MaxLedgerStateVersion);
```
dhedey, line 255: As mentioned, this shouldn't be a warning, as it's possible to request higher than the ndoe reports it has, courtesy of `GetFirstStateVersionToFetch`. I don't think we should even log as an Info. Maybe as Debug.
We could easily get this log multiple times a second in normal use due to one or more Core APIs lagging slightly.

src/RadixDlt.NetworkGateway.DataAggregator/Workers/NodeWorkers/NodeTransactionFetchWorker.cs
```diff
@@ -236,10 +238,31 @@ private async Task<FetchedTransactions> FetchTransactions(long fromStateVersion,
...
+            }
+            else
+            {
+                _logger.LogError(ex, "Unknown response error: {Message}", apiResponse.FailureResponse.Message);
```
dhedey, line 259: IMO we should just throw here instead of logging. The Worker's error logging and back-off logic will then kick in, which is what we want if we're hammering a Core API which is struggling.

dhedey (approved):
Approved, subject to a couple of requested tweaks.

src/RadixDlt.NetworkGateway.DataAggregator/Workers/NodeWorkers/NodeTransactionFetchWorker.cs
krzlabrdx: Changed to Debug + return.

src/RadixDlt.NetworkGateway.DataAggregator/Workers/NodeWorkers/NodeTransactionFetchWorker.cs
krzlabrdx: Agree.

dhedey (approved):

PR #508, opened 2023-09-27
Optimized ledger_transactions table usage

krzlabrdx:


src/RadixDlt.NetworkGateway.PostgresIntegration/Services/TransactionQuerier.cs
```diff
@@ -601,6 +601,7 @@ private async Task<ICollection<PendingTransactionSummary>> LookupPendingTransact
         GatewayModel.TransactionDetailsOptIns optIns,
         CancellationToken token)
     {
+        // TODO this query might possibly return tens of megabytes of data; depending on requested optIns we should include/exclude heavy columns
```
PawelPawelec-RDX, line 604: not sure we should leave that todo here.

PawelPawelec-RDX (approved):

PR #509, opened 2023-09-27
Dropped ORDER BY / LIMIT from pending TX payload queries

krzlabrdx:


dhedey (approved):
Nice fix. Should resolve the perf issue we see where the wrong index is used.

PR #534, opened 2023-10-11
Add new optin for receipt output

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -627,6 +627,13 @@ paths:
     post:
       operationId: ValidatorsUptime
       summary: Get Validators Uptime
+      description: |
+        Returns validators uptime data for time range limited by from_state_version and at_state_version.
```
krzlabrdx, line 631: Use backticks whenever we reference some property, i.e. "from_state_version" => "`from_state_version`"

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -627,6 +627,13 @@ paths:
...
+        `at_state_version` defines upper boundary (inclusive) for queried data. i.e `{ at_state_version = {epoch = 10} }`, will return uptime statistics from genesis till 10.
```
krzlabrdx, line 632: Shouldn't we document those properties directly in `ValidatorsUptimeRequest` definition? Properties "belong" to the request definition not endpoint definition in my opinion.

Endpoint docs:
Returns validator uptime for selected time range.

Request docs:
Returns... bla bla bla... 
from_state_version - lower time range
at_statE_version - upper time range
and so on and so forth

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -627,6 +627,13 @@ paths:
...
+        `from_state_version` defines lower boundary (inclusive) for queried data. i.e `{ from_state_version = {epoch = 10} }`, will return uptime statistics from epoch 10 till current max ledger tip.
+        it's possible to define both at once i.e `{ at_state_version = {epoch = 10}, from_state_version = {epoch = 25} }` will return uptime statistics for epochs from 10 till 25.
```
krzlabrdx, line 634: Ideally we should be very verbose on inclusiveness and exclusiveness when documenting ranges.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -627,6 +627,13 @@ paths:
...
+        you can use one of `state_version`, `epoch`, `epoch` and `round`, or `timestamp` to define from/at state version parameter, but then ongoing epoch will be selected and used for querying data.
+        i.e for request with `{ from_state_version = { state_version = 100 000}, at_state_version = { state_version = 300 000} }` gateway api will check in which epoch transactions with state version 100 000 and 300 000 and then use that as inclusive boundary for request.
```
krzlabrdx, line 636: 1. I have mixed feeling when it comes to pseudo-json format we used.
2. Same's true when it comes to the formatting of big numbers, maybe it's better to rewrite those samples and use `1000` to `3000` range?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1821,21 +1831,33 @@ components:
...
+          description: (false by default) if set to true, fee destination inside receipt object is returned.
           type: boolean
         receipt_costing_parameters:
+          description: (false by default) if set to true, costing parameters inside receipt object is returned.
           type: boolean
         receipt_events:
+          description: (false by default) if set to true, events inside receipt object is returned.
+          type: boolean
+        receipt_output:
+          description: (true by default) If set to true, transaction receipt output is returned.
```
krzlabrdx, line 1855: The default value of `true` should be considered a bug over here, isn't it? This is a collection of opt-INs after all. It would be super confusing for everyone to have this single opt-in default to `true` for whatever reason. Then again fixing this bug would break backwards compatibility.

@dhedey What's our policy with those issues? Is it still a bug or is it a feature by now?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1960,14 +1982,19 @@ components:
       type: object
       properties:
         ancestor_identities:
+          description: (false by default) if set to true, ancestor addresses - parent_address, owner_address and global_address for entities are returned.
```
krzlabrdx, line 1985: IMO it's kinda pointless to point out that default value of boolean property in optional object is `false` cause it's quite obvious in this context. However, if we really want to be that explicit in our docs I'd say we should move this information to the very back of the description, i.e. "(false by default) bla bla" => "bla bla. Defaults to `false`."

krzlabrdx: BTW use backticks whenever we reference a value, i.e. "false" => "`false`"

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1960,14 +1982,19 @@ components:
...
+          description: (false by default) if set to true, royalty vault balance for component entities is returned.
           type: boolean
         package_royalty_vault_balance:
+          description: (false by default) if set to true, royalty vault balance for package entities is returned.
           type: boolean
         non_fungible_include_nfids:
+          description: (false by default) if set to true, first page of non fungible ids are returned for each non fungible resource, with cursor which can be later used at `/state/entity/page/non-fungible-vault/ids` endpoint.
           type: boolean
         explicit_metadata:
+          description: (empty by default) allows specifying explicitly metadata properties which should be returned in response, limited to max 20 items.
```
krzlabrdx, line 1997: We should not mention values that are a subject to runtime configuration, such as max items per page. Those are not hardcoded values.

krzlabrdx (changes requested):

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: agreed on slack that unfrotunately that's the best option to do it this way.

krzlabrdx (approved):

PR #545, opened 2023-10-18
Optimized epoch and epoch + round based ledger state lookups

krzlabrdx:


src/RadixDlt.NetworkGateway.PostgresIntegration/DbQueryExtensions.cs
```diff
@@ -174,7 +174,8 @@ public static IQueryable<LedgerTransaction> GetLatestLedgerTransactionAtEpochRou
         return dbContext
             .LedgerTransactions
             .Where(lt => lt.Epoch == epoch && lt.RoundInEpoch >= round && lt.IndexInRound == 0)
-            .OrderByDescending(lt => lt.StateVersion)
+            .OrderByDescending(lt => lt.Epoch)
```
dhedey, line 177: Hmm I don't think this query is actually correct.

It says it "Returns most recently committed ledger transaction at or before given epoch and round." but then it filters to "lt => lt.Epoch == epoch && lt.RoundInEpoch >= round && lt.IndexInRound == 0".

Perhaps we should change the docs to "Returns the most recently committed round change ledger transaction at or before given epoch and round." and then change the where to just be on `lt.IndexInRound` ?

src/RadixDlt.NetworkGateway.PostgresIntegration/DbQueryExtensions.cs
```diff
@@ -192,7 +193,8 @@ public static IQueryable<LedgerTransaction> GetFirstLedgerTransactionAtEpochRoun
         return dbContext
             .LedgerTransactions
             .Where(lt => lt.Epoch >= epoch && lt.RoundInEpoch >= round && lt.IndexInRound == 0)
-            .OrderBy(lt => lt.StateVersion)
+            .OrderBy(lt => lt.Epoch)
+            .ThenBy(lt => lt.RoundInEpoch)
```
dhedey, line 197: Might be worth a comment that this should be fully specified due to the `lt.IndexInRound = 0`? (And associated partial index).

src/RadixDlt.NetworkGateway.PostgresIntegration/DbQueryExtensions.cs
dhedey: Okay, we've agreed that we probably want to change this at some point (likely to require an exact match) but for now, we'll ignore this logic and just fix the issue.

In which case - because we have a "where lt.Epoch == epoch" exact match, we don't need to order by it. We probably just want to order by descending on the round?

Or taking another step back, this should just equate to a direct look up of epoch and round?? So maybe we just do that?

src/RadixDlt.NetworkGateway.PostgresIntegration/DbQueryExtensions.cs
krzlabrdx: No longer relevant as we've dropped ORDER BY clause altogether - we rely on unique index anyways.

src/RadixDlt.NetworkGateway.PostgresIntegration/DbQueryExtensions.cs
krzlabrdx: Decided to go with exact epoch+round lookup.

PR #553, opened 2023-10-23
Fixed vault collection ordering for newly ingested data

krzlabrdx:


src/RadixDlt.NetworkGateway.PostgresIntegration/Models/EntityResourceVaultAggregateHistory.cs
```diff
@@ -84,6 +89,107 @@ internal class EntityResourceVaultAggregateHistory
...
+            ret.EnsureVaultSignificantUpdateStateVersionsExists();
+        }
+
+        return ret;
+    }
+
+    [MemberNotNull(nameof(VaultSignificantUpdateStateVersions))]
+    private void EnsureVaultSignificantUpdateStateVersionsExists()
+    {
+        VaultSignificantUpdateStateVersions ??= new List<long>(Enumerable.Repeat(1L, VaultEntityIds.Count));
```
dhedey, line 193: I wonder if we should eg use -1L or something here, to denote placeholder, rather than be confused for the actual state version 1?

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/EntityResourceVaultAggregateHistory.cs
```diff
@@ -84,6 +89,107 @@ internal class EntityResourceVaultAggregateHistory
...
+
+    private static EntityResourceVaultAggregateHistory CopyOrCreate(long databaseId, EntityResourceVaultAggregateHistory? other, long entityId, long resourceEntityId, long stateVersion)
+    {
+        var ret = new EntityResourceVaultAggregateHistory
+        {
+            Id = databaseId,
+            FromStateVersion = stateVersion,
+            EntityId = entityId,
+            ResourceEntityId = resourceEntityId,
+            VaultEntityIds = new List<long>(other?.VaultEntityIds.ToArray() ?? Array.Empty<long>()),
```
dhedey, line 176: This feels a little indirect? Could we not do `other?.VaultEntityIds.ToList() ?? List.Empty<long>()`?

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/EntityResourceVaultAggregateHistory.cs
```diff
@@ -84,6 +89,107 @@ internal class EntityResourceVaultAggregateHistory
...
+    /// </summary>
+    /// <returns>true if added or modified, otherwise false.</returns>
+    public bool TryUpsert(long vaultId, long stateVersion)
+    {
+        EnsureVaultSignificantUpdateStateVersionsExists();
+
+        var currentIndex = VaultEntityIds.IndexOf(vaultId);
+
+        // we're already the most recent one and there's no point potential second resource with the same state version
+        if (currentIndex == 0 && VaultEntityIds.Count == 1)
```
dhedey, line 128: These two if statements feel a bit arbitrary / overly optimized, and mean that `VaultSignificantUpdateStateVersions` isn't an accurate name - we don't update the state version of an item at the top of the list. Personally, I'd rather we just removed them both.

If we call it `__StateVersions`, I'd rather it was _accurately_ the state version of the last update. Otherwise, we should call it `RelativeUpdateOrdering` or something.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
```diff
@@ -1618,11 +1618,16 @@ void AggregateEntityResourceInternal(long entityId, long resourceEntityId)
...
+                    // we only want to create new aggregate resource value history entry if
+                    // - given vault is seen for the very first time,
+                    // - given vault is already stored but has been updated and this update caused change of order (this is evaluated right before db persistence)
+
                     var lookup = new EntityResourceVaultLookup(entityId, resourceEntityId);
 
                     if (mostRecentEntityResourceVaultAggregateHistory.TryGetValue(lookup, out var existingAggregate))
                     {
-                        if (existingAggregate.VaultEntityIds.Contains(resourceVaultEntityId))
+                        // we're already the most recent one, there's nothing more to do
```
dhedey, line 1629: Hmm this relates to my comments on the other file and the naming of last updated state version - which would make this early return incorrect.

Maybe a sensible performance improvement though 🤷 . I could be argued on this if we're willing to add a comment that it's not actually accurately the state versions.

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/EntityResourceVaultAggregateHistory.cs
```diff
@@ -84,6 +89,107 @@ internal class EntityResourceVaultAggregateHistory
...
+    }
+
+    public bool ShouldBePersisted()
+    {
+        if (_originalVaultEntityIds == null)
+        {
+            return true;
+        }
+
+        if (!_originalVaultEntityIds.SequenceEqual(VaultEntityIds.ToArray()))
```
dhedey, line 160: It feels like we should also check for a change in `VaultSignificantUpdateStateVersions`?

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/EntityResourceVaultAggregateHistory.cs
```diff
@@ -84,6 +89,107 @@ internal class EntityResourceVaultAggregateHistory
...
     public List<long> VaultEntityIds { get; set; }
+
+    /// <summary>
+    /// The last significant update state version for resources stored in <see cref="VaultEntityIds"/>.
+    /// </summary>
+    /// <remarks>
+    /// This property is nullable in order to ensure smooth database migration process.
+    /// </remarks>
+    [Column("vault_significant_update_state_versions")]
+    public List<long>? VaultSignificantUpdateStateVersions { get; set; }
```
dhedey, line 105: Out of interest - what's the purpose of storing this?

* It's inaccurate - or at least "Significant" needs spelling out, because it's taken a while to work out that it means "the last time it was updated and overtook some other resource as the most recently updated"
* I can't see anywhere it's actually used, other than for a minor optimization which is ignored anyway because of the SequenceEqual

Am I missing something?

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/EntityResourceVaultAggregateHistory.cs
krzlabrdx: There's no `List.Empty<long>` in C#. We could optionally move copy part to the if statement (`ret.VaultEntityIds.AddRange(other.ValueEntitiesIds)`)

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 Release Date: _unreleased_
 
 - Fixed `epoch [+ round]` based ledger state lookups.
+- Fixed vault collection ordering for newly ingested data. Database wipe might be required, see information below.
+
+*Warning* This release contains a fix for non-critical bug in data ingestion mechanism where resource vaults were not ordered correctly. Already ingested data will remain intact, i.e. bugfix will affect only newly ingested data.
```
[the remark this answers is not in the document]
krzlabrdx: changed.

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/EntityResourceVaultAggregateHistory.cs
dhedey: Ahh sorry, `other?.VaultEntityIds.ToList() ?? new List<long>()` - but this isn't a biggy :D

dhedey (approved):
Looks good, thanks 👍

PR #554, opened 2023-10-23
Added more strongly-typed OAS definition for programmatic_json

krzlabrdx:


src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1374,7 +1374,7 @@ components:
         emitter:
           type: object
         data:
-          type: object
```
dhedey, line 1377: Just a reminder (for want of a better place to write it) to add a description to the entity state endpoint explaining that eg the user may wish to cast to ProgrammaticScryptoSborValue.

dhedey (approved):
Looks good thanks - although just to flag the additional request regarding the state entity details documentation

PR #561, opened 2023-10-31
updated ephemeral test job jenkins jobs (plus some intellij auto-removal of whitespace)

les-sheppard:


les-sheppard:
No longer needed. Ephemeral environment deployments are being refactored so the they can be deployed by the testing harness.

PR #568, opened 2023-11-20
DO-2005 / fix caching behavior

kofrdx:
Set cache-suffix
Replace COPY . . 
Wait for joined images before deployment

The reason why `dotnet restore` and `dotnet build` are not cached is due to the `set-variables` action in setup-tags which modifies a file based on the .git hash of the latest commit. This seems to be relevant for displaying the version in the application.
We could consider removing this once the build time gets longer.


apps/DataAggregator/Dockerfile
```diff
@@ -16,7 +16,17 @@ RUN apt-get update -y \
 
 FROM mcr.microsoft.com/dotnet/sdk:7.0-bookworm-slim AS build
 WORKDIR /src
-COPY . .
+COPY sdk /src/sdk
```
krzlabrdx, line 19: 1. SDKs are not needed for Docker builds
2. I guess we could slightly improve readability by inlining multiple copy operations into single `COPY` call?

_(applies to all Dockerfiles)_

krzlabrdx (changes requested):

krzlabrdx (approved):

PR #575, opened 2023-11-29
return 408 status code on request timeout.

PawelPawelec-RDX:


krzlabrdx (approved):
Please add some info in CHANGELOG

PR #577, opened 2023-12-04
Update jenkins ci params

les-sheppard:


les-sheppard:
Using main branch instead of develop. Closing

PR #582, opened 2023-12-07
enable retries on transient db connection issues in gateway.

PawelPawelec-RDX:


PawelPawelec-RDX:
added retries to core api calls in gateway as well.

krzlabrdx (approved):

PR #585, opened 2023-12-12
transaction type and accounts with/without manifest owner call filters.

PawelPawelec-RDX:
That feature requires database rebuild or

`accounts_with_manifest_owner_method_calls` `accounts_without_manifest_owner_method_calls` filters and manifest classification will not work for transactions processed before deploy.

`accounts_without_manifest_owner_method_calls` filter will return all transactions even if owner method call was returned (because it's based on marker that's saved to db and that was coded in this pr, so transaction before deploying that feature will not have that marker)


PawelPawelec-RDX:
Waiting for Omar to finish RET implementation of extracting transaction type from manifest.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1656,6 +1656,18 @@ components:
     # Common Response Types
     #
 
+    TransactionType:
```
dhedey, line 1659: I think we're calling it TransactionClassification or TransactionCategory rather than "type"

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -2848,6 +2861,24 @@ components:
...
                 $ref: "#/components/schemas/StreamTransactionsRequestEventFilterItem"
+            accounts_with_manifest_owner_method_calls:
+              type: array
+              items:
+                $ref: "#/components/schemas/Address"
+            accounts_without_manifest_owner_method_calls:
+              type: array
+              items:
+                $ref: "#/components/schemas/Address"
+            transaction_type_filter:
```
dhedey, line 2872: Ditto - possibly `transaction_class_filter` ?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -2814,6 +2826,7 @@ components:
     # /stream/* Request & Response Types
     #
 
+    # TODO PP: update docs with examples.
```
dhedey, line 2829: Heads up - TODO still here - if it was intended to be done now? If not, we should raise a ticket for it and put the ticket reference here.

dhedey (commented):
Haven't had a chance for a full review - but have just had a quick schema review with a few comments

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: i wanted to complete it once it passes review :P I was kind of sure that name of props might change.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: i did go for `ManifestClass` which is how Omar called that in toolkit and I think it works better. What do you think?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: same as above I went for `manifest_class_filter`

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
```diff
@@ -286,14 +286,30 @@ private async Task<ExtendLedgerReport> ProcessTransactions(ReadWriteDbContext db
...
                         foreach (var address in extractedAddresses.All())
                         {
                             referencedEntities.MarkSeenAddress(address);
                         }
 
                         manifestExtractedAddresses[stateVersion] = extractedAddresses;
+
+                        var manifestSummary = toolkitManifest.Summary(_networkConfigurationProvider.GetNetworkId());
+
+                        for (var i = 0; i < manifestSummary.classification.Length; ++i)
```
krzlabrdx, line 301: Hint: do you know `++i` has literally no effect compared to regular `i++` in this case? :)

krzlabrdx (approved):

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
PawelPawelec-RDX: i know it has no difference in c#, i simply always use ++i:P

PR #587, opened 2023-12-13
effective fee calculation for validators.

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.PostgresIntegration/Services/ValidatorEffectiveFeeFactorProvider.cs
```diff
@@ -0,0 +1,102 @@
...
+
+namespace RadixDlt.NetworkGateway.PostgresIntegration.Services;
+
+public static class ValidatorEffectiveFeeFactorProvider
+{
+    public static GatewayModel.ValidatorCollectionItemEffectiveFeeFactor ExtractFeeFactorFromValidatorState(string validatorState, long currentEpoch)
+    {
+        var parsedState = JsonConvert.DeserializeObject<CoreModel.ValidatorFieldStateValue>(validatorState);
+
+        if (parsedState == null)
```
dhedey, line 78: I assume that `JsonConvert.DeserializeObject` doesn't throw then? It just returns null?

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/ValidatorEffectiveFeeFactorProvider.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            throw new UnreachableException($"Failed to parse state for validator. State: {validatorState}");
+        }
+
+        if (parsedState.ValidatorFeeChangeRequest == null)
+        {
+            return new GatewayModel.ValidatorCollectionItemEffectiveFeeFactor(
+                new GatewayModel.ValidatorCollectionItemEffectiveFeeFactorCurrent(parsedState.ValidatorFeeFactor));
```
dhedey, line 86: Very minor - I hate that the generated models use positional constructors with optional fields :D. To partially mitigate this, I tend to prefer using keyword arguments. Not sure if that's possible in this case?

Optional suggestion, feel free to ignore.

dhedey (approved):
Looks good, thanks

PR #590, opened 2023-12-20
return only assigned module's role assignments.

PawelPawelec-RDX:
that pull request requires database full resync, if not we'll return invalid data.


src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
```diff
@@ -406,6 +406,25 @@ nonFungibleResourceManagerFieldIdTypeSubstate.Value.NonFungibleIdType switch
                                 case CoreModel.ObjectTypeInfoDetails objectDetails:
                                     referencedEntity.PostResolveConfigure((ComponentEntity e) =>
                                     {
+                                        // TODO PP:
```
PawelPawelec-RDX, line 409: not sure about that.

krzlabrdx (approved):

PR #601, opened 2024-01-05
Reworked network configuration

krzlabrdx:


apps/GatewayApi/Controllers/StatusController.cs
```diff
@@ -118,7 +119,7 @@ public GatewayModel.NetworkConfigurationResponse NetworkConfiguration()
                 transactionProcessorPackage: wellKnownAddresses.TransactionProcessorPackage,
                 metadataModulePackage: wellKnownAddresses.MetadataModulePackage,
                 royaltyModulePackage: wellKnownAddresses.RoyaltyModulePackage,
-                accessRulesPackage: wellKnownAddresses.AccessRulesPackage,
+                accessRulesPackage: wellKnownAddresses.RoleAssignmentModulePackage,
```
PawelPawelec-RDX, line 122: hmmm shouldn't we rename that field ?:P I know it's breaking change but...

src/RadixDlt.NetworkGateway.Abstractions/Network/INetworkConfigurationProvider.cs
```diff
@@ -0,0 +1,182 @@
...
+
+public interface INetworkConfigurationProvider
+{
+    public Task<NetworkConfiguration> GetNetworkConfiguration(CancellationToken token = default);
+}
+
+public sealed class NetworkConfigurationProvider : INetworkConfigurationProvider
+{
+    private readonly CoreApi.StatusApi _statusApi;
+    // private readonly IEnumerable<INetworkConfigurationReaderObserver>? _oservers; // TODO restore
```
PawelPawelec-RDX, line 83: do we have task for these todos to not forget about them?

PawelPawelec-RDX (changes requested):

src/RadixDlt.NetworkGateway.Abstractions/Network/INetworkConfigurationProvider.cs
PawelPawelec-RDX: ok just reviewed second pr which addresses them. resolving.

apps/GatewayApi/Controllers/StatusController.cs
krzlabrdx: Yes, we definitely should! This change is part of separate PR (with xrd_usd_price)

PawelPawelec-RDX (approved):

PR #603, opened 2024-01-08
Core connectivity rework

krzlabrdx:


src/RadixDlt.NetworkGateway.DataAggregator/NodeServices/NodeWorkersRunner.cs
```diff
@@ -185,6 +183,8 @@ public bool IsHealthy()
     /// </summary>
     public async Task Initialize(CancellationToken cancellationToken)
     {
+        await Task.Yield();
```
PawelPawelec-RDX, line 186: if it's not async anymore maybe we can just change signature?

PawelPawelec-RDX (changes requested):

src/RadixDlt.NetworkGateway.DataAggregator/NodeServices/NodeWorkersRunner.cs
krzlabrdx: I'd prefer not to do that. Firstly I'd prefer not to touch the interface. Asynchronous capabilities for node initialization sound right to me. Moreover this method is actually used in `Task.WhenAll` as part of custom node pool: I'm not sure if this node pool will work correctly with synchronous node initialization. In other words: I'd prefer not to change anything in this area :)

PawelPawelec-RDX (approved):

PR #606, opened 2024-01-12
fix: load all non fungible resources

AbstractFruitFactory:


sdk/typescript/lib/subapis/state.ts
```diff
@@ -412,26 +463,59 @@ export class State {
...
+        fungible_resources: {
+          items: [
+            ...stateEntityDetails.fungible_resources.items,
+            ...results[0].fungible_resources.items,
+          ],
         },
-        non_fungible_resources: item.non_fungible_resources || {
-          total_count: 0,
-          items: [],
+        non_fungible_resources: {
```
dawidsowardx, line 512: we probably should include `total_count` field here?

dawidsowardx (approved):

PR #634, opened 2024-01-30
More verbose exceptions on node misconfiguration

krzlabrdx:


src/RadixDlt.NetworkGateway.Abstractions/Configuration/Extensions.cs
```diff
@@ -64,11 +64,45 @@
...
+using System.Collections.Generic;
+using System.Linq;
 
 namespace RadixDlt.NetworkGateway.Abstractions.Configuration;
 
 public static class Extensions
 {
+    public static List<CoreApiNode> GetEnabledNodes(this ICollection<CoreApiNode> nodes)
+    {
+        return nodes.Where(n => n.Enabled).ToList();
```
PawelPawelec-RDX, line 79: any reason we're not throwing here if 0 nodes are enabled? and shouldn't we have separate class responsible for node configuration validation?

krzlabrdx: I've decided not to change the current application logic where 0 enabled nodes are fine for some areas (e.g. healthchecks), even if this unavoidably leads to failure later on. Actually you're right. I've extend our existing configuration validation to ensure at least one Node is enabled.

src/RadixDlt.NetworkGateway.Abstractions/Configuration/Extensions.cs
```diff
@@ -64,11 +64,44 @@
...
+    public static CoreApiNode GetRandomEnabledNode(this ICollection<CoreApiNode> nodes)
+    {
+        return GetEnabledNodes(nodes).GetRandomBy(x => (double)x.RequestWeighting);
+    }
+
+    public static CoreApiNode GetRandomEnabledNodeForConstruction(this ICollection<CoreApiNode> nodes)
+    {
+        var availableNodes = GetEnabledNodes(nodes).Where(n => !n.DisabledForConstruction).ToArray();
+
+        if (!availableNodes.Any())
```
PawelPawelec-RDX, line 97: can we move that validation to `NetworkOptionsValidator` as well?

krzlabrdx: Not really as it is possible to run NG stack without those. What's more the exception should still remain as otherwise we'll encounter useless "sequence contains no elements".

PawelPawelec-RDX (approved):

PR #635, opened 2024-01-31
added CONTRIBUTING.md and CODE_OF_CONDUCT.md files.

PawelPawelec-RDX:
we keep tests in separate repository (test-harness) so it'd make sense to point ppl there and ask them to verify/write new tests and also regenerate client if they add any changes.


krzlabrdx (approved):

dhedey (approved):
Looks good! Can we also update the contributing section of the readme to point there?

PR #636, opened 2024-01-31
add PermanentlyRejectsAtEpoch to transaction status response.

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -2662,6 +2662,9 @@ components:
               type: integer
               format: int64
               nullable: true
+            permanently_rejects_at_epoch:
```
krzlabrdx, line 2665: 1. This should result in API version bump.
2. This is a property of a specific payload, i.e. it should belong to `known_payloads` collection items.
3. The description isn't super clear to me. I'd be more verbose in this case. Say "The epoch number at witch the transaction payload is guaranteed to get permanently rejected by the Network due to exceeded epoch range/validity" or something similar? We should make it clear this is some of guideline for the end client.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/PendingTransactionResponseAggregator.cs
```diff
@@ -0,0 +1,292 @@
...
+ * Work. You assume all risks associated with Your use of the Work and the exercise of
+ * permissions under this License.
+ */
+
+using RadixDlt.NetworkGateway.Abstractions.Model;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace RadixDlt.NetworkGateway.PostgresIntegration.Services;
```
krzlabrdx, line 70: Note for future us: if we decide to extract some logic to a separate file could we do that in two commits where the first one moves the logic as-is whereas another one adds/modifies whatever we're working on? Would be much easier to review.

krzlabrdx (changes requested):

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: 1. it's open discussion what API version is. 
2. are you sure it's not part of intent? i'm almost certain it is.
3. fixed.

krzlabrdx: 1. Yeah, there are some discussion in this area but still any API extension should result in version bump. It always did.
2. Oh, yes. My mistake!

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/PendingTransactionResponseAggregator.cs
```diff
@@ -0,0 +1,292 @@
...
+            GatewayApiSdk.Model.TransactionPayloadStatus.Pending =>
+                "This particular payload for this transaction has been accepted into a node's mempool on the network. It's possible but not certain that it will be committed to this network.",
+        };
+    }
+
+    private string GetIntentStatusDescription(GatewayApiSdk.Model.TransactionIntentStatus intentStatus)
+    {
+        return intentStatus switch
+        {
+            GatewayApiSdk.Model.TransactionIntentStatus.Unknown =>
```
krzlabrdx, line 276: Please use `GatewayModel = RadixDlt.GatewayApiSdk.Model` to keep the codebase consistent.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/PendingTransactionResponseAggregator.cs
```diff
@@ -0,0 +1,292 @@
...
+            GatewayApiSdk.Model.TransactionIntentStatus.Unknown => GatewayApiSdk.Model.TransactionStatus.Unknown,
+            GatewayApiSdk.Model.TransactionIntentStatus.CommittedSuccess => GatewayApiSdk.Model.TransactionStatus.CommittedSuccess,
+            GatewayApiSdk.Model.TransactionIntentStatus.CommittedFailure => GatewayApiSdk.Model.TransactionStatus.CommittedFailure,
+            GatewayApiSdk.Model.TransactionIntentStatus.CommitPendingOutcomeUnknown => GatewayApiSdk.Model.TransactionStatus.Pending,
+            GatewayApiSdk.Model.TransactionIntentStatus.PermanentlyRejected => GatewayApiSdk.Model.TransactionStatus.Rejected,
+            GatewayApiSdk.Model.TransactionIntentStatus.LikelyButNotCertainRejection => GatewayApiSdk.Model.TransactionStatus.Pending,
+            GatewayApiSdk.Model.TransactionIntentStatus.Pending => GatewayApiSdk.Model.TransactionStatus.Pending,
+        };
+
+        var statusesToReturnPermanentlyRejectsAtEpoch = new[]
```
krzlabrdx, line 230: This should be a static readonly value, not a variable.

krzlabrdx (changes requested):

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/PendingTransactionResponseAggregator.cs
PawelPawelec-RDX: well it's the only place where it's used. separating it from usage will only make you jump between two places and harder to read the code. I see no benefit from doing that.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: 1. well it should result but we always matched it with release version, so in this case that would be 2.0.0

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/PendingTransactionResponseAggregator.cs
krzlabrdx: This is static, constant value, not a variable. Feel free to define it as static within the narrowest possible scope. We don't have to allocate it on every single usage.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1662,6 +1658,13 @@ components:
     # Common Response Types
     #
 
+    Epoch:
+      nullable: true
```
krzlabrdx, line 1662: Ah, damn it... this is invalid. Or rather... this is invalid name if it is meant to be nullable type. Right now we can't use this type where epoch is a mandatory property. Maybe let's just copy this definition to those two types.

PR #637, opened 2024-01-31
Ephemeral env test job run for all branches

Theoklitos:


deployment/ci.env
```diff
@@ -1,4 +1,4 @@
 # Git commit version of most recent Node compatible with this Network Gateway (only used in CI/CD environments)
-FULLNODE_COMMIT=fd4f682acacdcfc52c98ff019db8f16051774230
-FULLNODE_VERSION=sha-fd4f682
+FULLNODE_COMMIT=65ddf94b2372e6bada89f5d239088647b6d5c7c5
+FULLNODE_VERSION=release-anemone
```
PawelPawelec-RDX, line 3: can we point to fixed image version? we don't expect `release-anemone` but in general we should use sha based iamges. if possible (newest from main is `sha-72d26e5`)

PawelPawelec-RDX (approved):

deployment/ci.env
Theoklitos: done

PR #646, opened 2024-02-06
Release/babylon 1.4.0 - > main

PawelPawelec-RDX:


krzlabrdx (approved):

krzlabrdx (approved):

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -879,11 +879,99 @@ tags:
             }
       ```
 
+      ### Iterating over all key value store keys
+
+      You can use `/state/key-value-store/keys` endpoint to iterate over all keys in certain key value store.
```
dhedey: Minor English article corrections:
```
      You can use the `/state/key-value-store/keys` endpoint to iterate over all the keys in a certain key value store.
```

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -879,11 +879,99 @@ tags:
...
+      If the queried key value store contains more than one page of keys, a next_cursor will be returned, which can be used as cursor in the request to fetch further pages.
```
dhedey: Minor English article corrections:
```
If the queried key value store contains more than one page of keys, a `next_cursor` will be returned, which can be used as a cursor in the request to fetch further pages.
```

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -879,11 +879,99 @@ tags:
...
+      i.e:
+      ```
+      /state/key-value-store/keys
+
+      {
+        "key_value_store_address": "internal_keyvaluestore_tdx_2_1kzjd929eqlzd9n02uuj8jd48705vcrpvhv4mgxnaltrgystnca3qxk"
+      }
+      ```
+
+      Will respond with list of all keys (in both `hex` and `programmatic_json` form) and state version when they were last updated.
```
dhedey, line 896: Minor English article corrections:

```
      Will respond with a list of all the keys (in both `hex` and `programmatic_json` form) and the state version when they were last updated.
```

However(!) I'd suggest we say it's actually not all keys - it's importantly the *first page of keys*, and we should explain how it's ordered too.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -879,11 +879,99 @@ tags:
...
+
```
dhedey, line 897: Also - just to verify - if a key value store entry is **deleted**, what happens?

I think it's still returned by this API? Or do we filter them out?

If it doesn't filter them out, we should explain that this API doesn't do so, and explain how the entry read endpoint handles us looking up a key of an entry that's been deleted.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -879,11 +879,99 @@ tags:
...
+          ...
+        ],
+        "key_value_store_address": "internal_keyvaluestore_tdx_2_1kzjd929eqlzd9n02uuj8jd48705vcrpvhv4mgxnaltrgystnca3qxk"
+      }
+      ```
+
       ### Querying for the content under specific keys
 
-      After that you can use `/state/key-value-store/data` endpoint to query the content for given keys. To do that you can either use key in form of `json` or `hex`.
+      You can either rely on keys returned from `/state/key-value-store/keys` endpoint or construct keys in json form on your own. It's possible if you created that key value store and you know what is the form of keys.
```
dhedey: Suggested:
```
You can either use the keys returned from the `/state/key-value-store/keys` endpoint or construct keys in json form on your own, if you know the key type of the `KeyValueStore`.
```

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -879,11 +879,99 @@ tags:
...
+
+      #### Querying using hex keys
+
+      Simply grab `raw_hex` key returned from `/state/key-value-store/keys` endpoint and call `/state/key-value-store/data` with it as parameter.
```
dhedey, line 925:
```
      Simply grab the `raw_hex` key returned from the `/state/key-value-store/keys` endpoint and call `/state/key-value-store/data` with it as a parameter.
```

dhedey (commented):
Awesome, thanks Pawel!

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: how about

```
Will respond with a first page of all keys (in both `hex` and `programmatic_json` form) and the state version when they were last updated. Returned keys are ordered by their last change, first key on a list will be one that was most recently updated/created. 
```

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: we filter out deleted keys. do we want to document that somehow?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
dhedey: I suggest this tweak, if you're happy with it? 
```
Will respond with a first page of the keys of the current entries of the key value store. This collection is ordered by the state version of the entry's last update, descending. In other words, the first entry on the list will be one that was most recently updated/created. Each item includes the entry's key (in both `hex` and `programmatic_json` form) and the state version when it was last updated.
```

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
dhedey: I think if we rewrite as per my other message, and say "current entries", this implies deleted keys aren't returned.

PR #651, opened 2024-02-12
update docs after David's review.

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -194,6 +196,115 @@ tags:
       }
       ```
 
+      ## Consistent reads
+
+      As described in [Browsing historical data](#section/Browsing-historical-data) in all Network Gateway requests it is possible to specify the exact ledger state by providing the `at_ledger_state` parameter.
```
krzlabrdx, line 201: This is not 100% true. Could we please clarify that this is applicable to all `/state/*` and `/stream/*` endpoints and in case of other areas it should be read from endpoint docs?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -194,6 +196,115 @@ tags:
...
+          "epoch": 74324,
+          "round": 84
+        },
+        "total_count": 128,
+        "next_cursor": "eyJvIjoxMDB9",
+        "items": [...]
+      }
+      ```
+
+      2nd page of keys:
```
krzlabrdx, line 234: Might be useful to be super verbose here, maybe we should add some hint such as: _pay attention to the atLedgerState.StateVersion - it matches the state version from previous (initial) request_. What do you think?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -194,6 +196,115 @@ tags:
...
+
+      Response:
+      ```
+      {
+        "ledger_state": {
+          "network": "mainnet",
+          "state_version": 50576725,
+          "proposer_round_timestamp": "2024-02-20T16:46:22.343Z",
+          "epoch": 74324,
+          "round": 84
```
krzlabrdx, line 226: Just thinking out loud here. I'd trim some of those unimportant response properties (`netwrok, timestamp, epoch, round`) and leave only those that are important to understand the concept. Just to reduce the noise in those examples. Similarly to how we `...` all the `items`.

krzlabrdx (approved):

dhedey (approved):

PR #653, opened 2024-02-13
Fixed unstable package blueprint and code aggregation

krzlabrdx:


src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/EmptyDictionary.cs
```diff
@@ -0,0 +1,9 @@
+using System.Collections.Generic;
+
+namespace RadixDlt.NetworkGateway.PostgresIntegration.LedgerExtension;
+
+internal static class EmptyDictionary<TKey, TValue>
```
PawelPawelec-RDX, line 5: Do we need that extension? what about using `ImmutableDictionary<TKey,TValue>.Empty` ?

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/EntityMetadataProcessor.cs
```diff
@@ -0,0 +1,214 @@
...
+                {
+                    var currentPosition = aggregate.MetadataIds.IndexOf(previousEntry.Id);
+
+                    if (currentPosition != -1)
+                    {
+                        aggregate.MetadataIds.RemoveAt(currentPosition);
+                    }
+                }
+
+                // TODO introduce entry.IsDeleted extension method
```
PawelPawelec-RDX, line 103: do we have any task for that? can we maybe add it to this pullrequest? I'm affraid that we'll just leave that todo here forever, so either
1. create a task for that
2. code it in this pr
3. remove todo.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/ChangeTracker.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        return _store.GetOrAdd(key, _ =>
+        {
+            _insertionOrder.Add(key);
+
+            return factory(key);
+        });
+    }
+
+    public IEnumerable<KeyValuePair<TKey, TValue>> AsEnumerable()
```
PawelPawelec-RDX, line 25: Do we need to use enumrable? I'm not big fan of that interface, can we use some other interace i.e IReadOnlyCollection or some concrete implementation?

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/EmptyDictionary.cs
krzlabrdx: I knew there was something built-in but I somehow couldn't find it! Yes, obviously we should use `ImmutableDictionary<TKey,TValue>.Empty`.  It actually even naturally pushes the code into what I'd imagine it to look like long-term.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/ChangeTracker.cs
krzlabrdx: Enumerable is a right choice here as it does **not** result in needless memory allocation. Just not being fan of something isn't enough here in my opinion. ;-)

krzlabrdx: Ah, and I forgot about the most important one. This interface protects us against concurrent modifications. An attempt to `.GetOrAdd` while iterating will result in regular `Collection was modified; enumeration operation may not execute.` exception just as with regular Dictionary. This is important.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/EntityMetadataProcessor.cs
krzlabrdx: I've literally moved as much existing code as possible. I don't think I'd like to change every todo right now.

PawelPawelec-RDX: we had that conversation already :P we decided to either 
1. track these todo by jira task
2. leave that todo, if you're 100% sure that it'll be addressed in very short term. (not a big fan of that option though, it should never hit develop branch and wip tasks can be tracked other way)
3. remove it

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/ChangeTracker.cs
PawelPawelec-RDX: well I don't see any place where we modify that so we're safe with using `IReadOnlyCollection`. and i'm not a fan of `IEnumerable` because it leads into potentially hard to detect errors when it's hard to track when and where collection has changed. In that example it's not an issue but i'd like to avoid that interface if possible. Feel free to ignore that comment if you strongly disagree.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/EntityMetadataProcessor.cs
krzlabrdx: Removed it.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/ChangeTracker.cs
krzlabrdx: > well I don't see any place where we modify that so we're safe with using IReadOnlyCollection.
The idea behind the code is to make sure it will throw the moment we start doing something incorrectly. 

> and i'm not a fan of IEnumerable because it leads into potentially hard to detect errors [...]
Concurrent dictionary modifications during iteration are even more difficult to detect/spot/fix. 

> Feel free to ignore that comment if you strongly disagree.
I have a rather strong opinion this time. I'm going to ignore the comment then.

PawelPawelec-RDX (approved):

PR #656, opened 2024-02-16
Component royalties

krzlabrdx:


CHANGELOG.md
```diff
@@ -13,9 +13,12 @@ Release Date: _not published yet_
 ### API Changes
 - Added new filter `manifest_badges_presented_filter` to `/stream/transactions` endpoint which allows filtering transactions by badges presented.
 - Added `/state/key-value-store/page/keys` and `/state/non-fungible/page/ids` endpoints to replace now obsoleted `/state/key-value-store/keys` and `/state/non-fungible/ids`.
+- Added strong type definition of `royalty_config` for package blueprints and general components.
```
PawelPawelec-RDX, line 16: hmmm it's not really API Change, we're still returning same data there, maybe we should move that entry somewhere else?

PawelPawelec-RDX (commented):
looks good but:

1. there are some tests failing 
2. make sure to write some tests for these features

CHANGELOG.md
krzlabrdx: This is definitely a change, some could argue it's even a breaking change. ;-) Changelog is meant to keep track of notable changes to the API / application. This change impacts API users / SDK users. IMO it should be noted.

krzlabrdx: The schema (returned data) is obviously one of the primary API design aspects but it is not the only one, and we cannot forget abut those other aspects in our changelog. I'd keep it the way it is to avoid unnecessary changelog fragmentation. Should we end up in a situation where we have a dozen of "actual schema changes" and a handful of "schema unrelated changes to the API specs" I guess we should consider this division between the two. Do you agree?

PawelPawelec-RDX (approved):

PR #670, opened 2024-03-04
fix order of processed changes (dictionary doesn't guarantee order), …

PawelPawelec-RDX:
…make sure we create new value per each element (bug with updating previous reference).


src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/KeyValueStoreProcessor.cs
```diff
@@ -0,0 +1,281 @@
...
+
+internal record struct KeyValueStoreChangePointerLookup(long KeyValueStoreEntityId, long StateVersion, ValueBytes Key);
+
+internal record KeyValueStoreChangePointer(ReferencedEntity ReferencedEntity, CoreModel.GenericKeyValueStoreEntrySubstate KeyValueStoreEntry);
+
+internal class KeyValueStoreProcessor
+{
+    private readonly ProcessorContext _context;
+
+    private Dictionary<KeyValueStoreChangePointerLookup, KeyValueStoreChangePointer> _changePointers = new();
```
krzlabrdx, line 86: Please use `ChangeTracker` instead of `Dictionary+List`.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/KeyValueStoreProcessor.cs
```diff
@@ -0,0 +1,281 @@
...
+            var isDeleted = change.KeyValueStoreEntry.Value == null;
+            var entry = new KeyValueStoreEntryHistory
+            {
+                Id = _context.Sequences.KeyValueStoreEntryHistorySequence++,
+                KeyValueStoreEntityId = lookup.KeyValueStoreEntityId,
+                FromStateVersion = lookup.StateVersion,
+                Key = change.KeyValueStoreEntry.Key.KeyData.GetDataBytes(),
+                IsLocked = change.KeyValueStoreEntry.IsLocked,
+                IsDeleted = isDeleted,
+                Value = isDeleted ? null : change.KeyValueStoreEntry.Value!.Data.StructData.GetDataBytes(),
```
krzlabrdx, line 161: Absolutely minor notice: why do we need `!.` if we already know it is not null (`isDeleted == false`)?

krzlabrdx (changes requested):

krzlabrdx (approved):

PR #686, opened 2024-03-17
Fix configuration for page size in compose file and environment template

Krulknul:
I was having issues trying to set up a different maximum page size for my own gateway instance. 

I would make the maximum higher in the `.env`, but it would still give an error when I tried to request more than 100 items in a page. I asked around on Discord and Mleekko came up with the answer [here](https://discord.com/channels/417762285172555786/1149370695206318151/1218902879813894234). Seems somehow the docker-compose file was incorrect.

I tested Mleekko's answer by setting the `GatewayApi__Endpoint__MaxPageSize` to 5. This caused the gateway to crash, as this variable may not be lower than the `GatewayApi__Endpoint__DefaultPageSize`. That's why I also added this to the docker-compose file and the environment template.


PawelPawelec-RDX:
thanks! nice spot. Can you please create pr against develop branch? @Krulknul

Krulknul:
> thanks! nice spot. Can you please create pr against develop branch? @Krulknul

Ah, Excuse me. Should have read the contribution document. I'm on it.

PR #687, opened 2024-03-18
Fix configuration for page size in compose file and environment template

Krulknul:
Duplicate of #686 but against developer branch.


PawelPawelec-RDX (approved):

PawelPawelec-RDX:
Thanks a lot for your contribution!

PR #703, opened 2024-03-29
Fix processing authorized badges if badge doesn't exist and or used i…

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/AccountAuthorizedDepositorsProcessor.cs
```diff
@@ -75,7 +75,7 @@
 
 namespace RadixDlt.NetworkGateway.PostgresIntegration.LedgerExtension;
 
-internal record struct AccountAuthorizedDepositorsDbLookup(long AccountEntityId, long ResourceEntityId, long? NonFungibleIdDataId);
+internal record struct AccountAuthorizedDepositorsDbLookup(long AccountEntityId, long ResourceEntityId, string? NonFungibleId);
```
krzlabrdx, line 78: This and the correlated SQL query might blow into our faces later on. Maybe we should treat both variants separately from day one?

krzlabrdx (approved):

PR #704, opened 2024-03-29
Transaction pre validation endpoint

PawelPawelec-RDX:
There are many todos to fill in description, i'll fill that once we agree how contract should look like.


src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -2387,7 +2411,90 @@ components:
...
+    AccountDepositPreValidationDecidingFactors:
+      description: Deciding factors used to calculate response.
+      type: object
+      required:
+        - default_deposit_rule
+      properties:
+        is_badge_authorized_depositor:
+          description: Whether the input badge belongs to the account's set of authorized depositors. This field will only be present if any badge was passed in the request.
+          type: boolean
+          nullable: true
```
krzlabrdx, line 2433: I think it's enough to not enlist this type in `required`?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -2387,7 +2411,90 @@ components:
...
+
+    TransactionAccountDepositPreValidationAuthorizedDepositorBadge:
+      type: object
+      required:
+        - badge_type
+        - resource_address
+      properties:
+        badge_type:
+          $ref: "#/components/schemas/AccountAuthorizedDepositorBadgeType"
+        resource_address:
```
krzlabrdx, line 2477: I believe we were supposed to move `resource_address` to derived types as the fact that it exists in all derived types today is nothing more but a coincidence.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -3011,6 +3118,44 @@ components:
...
+              $ref: "#/components/schemas/Address"
+            resource_addresses:
+              type: array
+              items:
+                $ref: "#/components/schemas/Address"
+            badge:
+              $ref: "#/components/schemas/TransactionAccountDepositPreValidationAuthorizedDepositorBadge"
+          example:
+            account_address: "<account_address>"
+            resource_addresses: ["<resource_address>"]
```
krzlabrdx, line 3138: IIRC we use hyphens not underscores for those placeholders?

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/QueryHelper.cs
```diff
@@ -0,0 +1,110 @@
...
+    internal static async Task<TEntity> GetEntity<TEntity>(ReadOnlyDbContext dbContext, EntityAddress address, GatewayApiSdk.Model.LedgerState ledgerState, CancellationToken token)
+        where TEntity : Entity
+    {
+        var entity = await dbContext
+            .Entities
+            .Where(e => e.FromStateVersion <= ledgerState.StateVersion)
+            .AnnotateMetricName()
+            .FirstOrDefaultAsync(e => e.Address == address, token);
+
+        if (entity == null)
```
krzlabrdx, line 98: Just remember that we can't use it anywhere virtual accounts/identities are expected! Are we OK with that?

krzlabrdx (changes requested):

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: it defaulted to false without it. I think it handles differently simple types.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: yes, but couldn't force openAPI generator to properly generate classes if that was inside child classes.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/QueryHelper.cs
PawelPawelec-RDX: yes, we don't expect virtual entities in places where I've used it.

krzlabrdx (approved):

PR #705, opened 2024-04-02
do not add lower state version filter if lower state version is null.

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.PostgresIntegration/Services/TransactionQuerier.cs
```diff
@@ -141,10 +141,21 @@ public async Task<TransactionPageWithoutTotal> GetTransactionStream(TransactionS
             ? request.Cursor?.StateVersionBoundary ?? request.FromStateVersion
             : request.FromStateVersion;
 
-        var searchQuery = _dbContext
-            .LedgerTransactionMarkers
-            .Where(lt => lt.StateVersion <= upperStateVersion && lt.StateVersion >= (lowerStateVersion ?? lt.StateVersion))
-            .Select(lt => lt.StateVersion);
+        IQueryable<long> searchQuery;
+        if (lowerStateVersion.HasValue)
```
krzlabrdx, line 145: While the idea to completely eliminate lower state version check in case it's not needed is certainly a good idea, this implementation isn't. We're doubling way too much code, it's definitely error-prone.

```
quuery = ....
  .Where(lt => lt.StateVersion <= upperStateVersion)
  .ApplyLowerStateVersion(lowerStateVersion)
  ...
ApplyLowerStateVersion(this IQuerable<LedgerTxMarkers> searchQuery, long? lowerStateVersion)
{
  return lowerStateVErsion.HasValue ? searchQuery.Where(ltm => ltm.StateVersion >= lowerStateVErsion.Value) : searchQuery;
} 
```

krzlabrdx (changes requested):

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/TransactionQuerier.cs
PawelPawelec-RDX: are you sure it's gonna be translated well into sql by entity framework and not executed in memory after loading all the data? 

but yes we can try to rewrite it differently so there's way less code duplication, I just created draft while in rush yesterday

PR #707, opened 2024-04-03
config value for min batch size.

PawelPawelec-RDX:


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Release built: _not published yet_
   - Dockerfiles no longer specify custom `app` user as it comes built-in with official base images.
   - Removed now-obsolete or no-longer-needed code.
   - Prometheus integration exposes new built-in metric `httpclient_request_duration_seconds_bucket` for all registered HTTP client. 
+- Introduced new configuration parameter `LedgerConfirmation.MinCommitBatchSize`. It controls the minimum batch size of transactions that will be processed, set by default to 1.
```
krzlabrdx, line 20: I'd probably suggest not to announce that as it turned out it's going ot be rather "internal feature".

krzlabrdx (approved):

src/RadixDlt.NetworkGateway.DataAggregator/Services/FetchedTransactionStore.cs
```diff
@@ -161,15 +161,22 @@ public List<CoreApiSdk.Model.CommittedTransaction> GetTransactionBatch(long from
...
+        var transactions = nodeWithMostTransactions
             .Where(x => x.Key >= fromStateVersion)
             .OrderBy(x => x.Key)
             .Take(maxBatchSize)
             .Select(x => x.Value.CommittedTransaction)
             .ToList();
+
+        if (transactions.Count < minBatchSize)
+        {
+            return [];
```
dhedey, line 176: I fear this won't work well if there is a large min batch size -- because the node has a dynamic upper batch size dependent on how large the response is. This is to protect the Node and Gateway from an out-of-memory if the response is too large.

This code as is could get us stuck, if, `ShouldFetchNewTransactions` is false after 4 transactions, but our min batch size is 5.

Instead, we should probably also check that ShouldFetchNewTransactions is true for that node name before returning `[]` - else we should just handle the transactions as-is.

We should also comment this I imagine!

src/RadixDlt.NetworkGateway.DataAggregator/Configuration/LedgerConfirmationOptions.cs
```diff
@@ -101,6 +101,13 @@ public sealed record LedgerConfirmationOptions
     [ConfigurationKeyName("MaxCommitBatchSize")]
     public long MaxCommitBatchSize { get; set; } = 300;
 
+    /// <summary>
+    /// Gets or sets MinCommitBatchSize.
+    /// The minimum batch to send to the ledger extension service for committing.
```
dhedey, line 106: We should probably explain that this is mostly exposed for testing, and we expect 1 is the best value for production use.

dhedey (changes requested):
I've noticed a potential issue and flagged a possible workaround

src/RadixDlt.NetworkGateway.DataAggregator/Services/FetchedTransactionStore.cs
PawelPawelec-RDX: added explicit exception informing that gateway is missconfigured and will not process transaction in such config.

src/RadixDlt.NetworkGateway.DataAggregator/Configuration/LedgerConfirmationOptions.cs
PawelPawelec-RDX: done

CHANGELOG.md
PawelPawelec-RDX: removed

PR #714, opened 2024-04-10
Account Locker endpoints proposal

krzlabrdx:
Two new endpoints:
1. Paginate through all the resources of a given account+account-locker tuple.
2. Find if a given account+account-locker tuple has been altered at all with optional `from_state_version` (useful for clients that might want to cache this information)


src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1719,6 +1719,52 @@ paths:
...
+        - State
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/WalletTbdRequest"
+      responses:
+        "200":
+          description: Validators Uptime
```
dhedey, line 1760: nitpick: TBD :D

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1719,6 +1719,52 @@ paths:
...
+        "200":
+          description: TBD
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/StateAccountLockerPageAccountResourcesResponse"
+        "4XX":
+          $ref: "#/components/responses/ErrorResponse"
+
+  /wallet/tbd: # TODO ok, we need to figure out suitable naming scheme, and then rename some types
```
dhedey, line 1745: :+1: I'll ignore the WalletTbd prefix

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -4224,6 +4299,37 @@ components:
...
+        - $ref: "#/components/schemas/CursorLimitMixin"
+        - type: object
+          required:
+            - account_locker_address
+            - account_address
+          properties:
+            account_locker_address:
+              $ref: "#/components/schemas/Address"
+            account_address:
+              $ref: "#/components/schemas/Address"
```
dhedey, line 4314: We're missing optional `resource_addresses` (and I assume a good description of the two modes is to come on the endpoint itself)

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -2113,6 +2159,35 @@ components:
             typed:
               $ref: "#/components/schemas/MetadataTypedValue"
 
+    AccountLockerResourceVaultCollection:
+      description: TBD
+      allOf:
+        - $ref: "#/components/schemas/ResultSetCursorMixin"
```
dhedey, line 2165: I assume we will be able to provide sensible values here if a user specifies `resource_addresses`?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -4273,6 +4379,64 @@ components:
             validators:
               $ref: "#/components/schemas/ValidatorUptimeCollection"
 
+    #
+    # /wallet/* Request & Response Types
+    #
+
+    WalletTbdRequest:
+      allOf:
+        - $ref: "#/components/schemas/AtLedgerStateMixin" # TODO do we even want to support that? if yes, it could technically fall under regular /state/* endpoints
```
dhedey, line 4388: Yeah I think we can support this relatively easily if we store the fully history. I think we may as well just store the full history.

But I don't think it's necessary a `/state/*` endpoint conceptually.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -4273,6 +4379,64 @@ components:
...
+        - type: object
+          required:
+            - lookup
+          properties:
+            lookup:
+              type: array
+              items:
+                $ref: "#/components/schemas/WalletTbdRequestLookupTuple"
+      example:
```
dhedey, line 4397: nitpick: Example is incorrect

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -4273,6 +4379,64 @@ components:
...
+            items:
+              type: array
+              items:
+                $ref: "#/components/schemas/WalletTbdResponseItem"
+    WalletTbdResponseItem:
+      type: object
+      required:
+        - account_locker_address
+        - account_address
+        - last_updated_at_state_version
```
dhedey, line 4430: I don't think we should call it `last_updated_at_state_version`. I think it should perhaps be called `resource_last_stored_at_state_version` - and also it needs to be optional - if an `(account_locker_address, account_address)` tuple has never had any resources stored (which will be the most common case).

Or do you intend to not return such a state?

I think I'd personally prefer returning an option for each requested address, so it can be zipped with the request - but if you feel strongly for us not to, that's okay. Either way, we should document the endpoint in the description.

dhedey (commented):
Thanks for pulling this together - I think such a draft PR is the easiest way to discuss these proposals! 👍

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
krzlabrdx: 1. Renamed.
2. The idea is not to return "missing" tuples. This is the model we use elsewhere. What's more it reduces number of null checks / unnecessarily nullable properties.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
krzlabrdx: TBD'ed

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
krzlabrdx: I'll leave it to future us to decide if we want to support that or not.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
krzlabrdx: Descriptions and actual endpoint paths are beyond the scope of the schema proposal.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
krzlabrdx: Yeah, should the client specify `explicit_resource_addresses` we'll return responses for those that were requested.

dhedey: Sorry my comment wasn't very clear. I meant to the `ResultSetCursorMixin` - e.g. no cursor, size, or whatever it includes?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
dhedey: We should decide now I think to make implementation run smoothly. Let's say "yes" unless you object strongly.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
dhedey: > This is the model we use elsewhere. What's more it reduces number of null checks / unnecessarily nullable properties.

Can I just ask for "where" it's used elsewhere when we have an explicit batch lookup? Is that our approach in e.g. metadata?

For endpoints in the new sub-api, I'd rather we go back to the drawing board and use the pattern we think is best (easiest to use for a client and most flexible for extension or changes in the backend). I think the zippable approach wins on both fronts.

If it makes it conceptually nicer, I think we should even consider this kind of format `{ key: ObjectX, value: ObjectY? }` for these kinds of explicit lookup requests against some index.

In this case, we'd get structures like this: `{ key: { account_locker_address, account_address }, value: None }` and `{ key: { account_locker_address, account_address }, value: { last_updated_at_state_version } }`

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
krzlabrdx: Yeah, it should work just fine. However please note that in theory we could expect configure max_page_size = 10 and max_explicit_resource_address = 50 and then you'd have to use cursor to iterate over 5 pages of explicitly requested resources. :)

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -4273,6 +4383,62 @@ components:
...
+              items:
+                $ref: "#/components/schemas/WalletTbdResponseItem"
+    WalletTbdResponseItem:
+      type: object
+      required:
+        - account_locker_address
+        - account_address
+        - resource_last_stored_at_state_version
+      properties:
+        account_locker_address:
```
dhedey, line 4434: Decision: let's create separate key and value objects here, where `value` is optional and contains the required resource_last_stored_at_state_version - and we will return an entry for each key.

dhedey: Name of API will be decided in slack thread. I'd like to get that resolved this week. Russ to give sign-off if you need a go/no-go.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1824,6 +1824,56 @@ paths:
...
+        - State
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/StateAccountLockerPageVaultsRequest"
+      responses:
+        "200":
+          description: Account locker account resources (paginated)
```
PawelPawelec-RDX, line 1844: vaults?

PawelPawelec-RDX (approved):

PR #731, opened 2024-04-25
Dropped max_page_size validation

krzlabrdx:


CHANGELOG.md
```diff
@@ -42,7 +42,8 @@ Release built: _not published yet_
 - Added new endpoint `/state/package/page/blueprints` returning paginable iterator over package blueprints.
 - Added new endpoint `/state/package/page/codes` returning paginable iterator over package codes.
 - Added new endpoint `/state/entity/page/schemas` returning paginable iterator over entity schemas.
-- Added new endpoint `/transaction/account-deposit-pre-validation` which allows to pre-validate if deposits can succeed based on account deposit settings and badges presented, before submitting the transaction. 
+- Added new endpoint `/transaction/account-deposit-pre-validation` which allows to pre-validate if deposits can succeed based on account deposit settings and badges presented, before submitting the transaction.
+- `limit_per_page` request parameter is no longer validated against `*MaxPageSize` API configuration parameters. In case requested limit exceeds API configuration maximum value is used. This change is meant to reduce clients need to understand and honor API configuration.
```
PawelPawelec-RDX, line 46: can we also fix documentation in schema file? we state there

`If the `limit_per_page` request parameter is lower than the number of transactions to be returned, a `next_cursor` will be returned. It can be used in subsequent requests as the `cursor` parameter to fetch further pages.
`

which will no longer be true.

We should also mention our logic change in both changelog and docs, that smaller page size might be returned if it's higher than limit.

src/RadixDlt.NetworkGateway.GatewayApi/Configuration/EndpointOptions.cs
```diff
@@ -106,6 +106,22 @@ public sealed class EndpointOptions
 
     [ConfigurationKeyName("TransactionStreamMaxFilterCount")]
     public int TransactionStreamMaxFilterCount { get; set; } = 10;
+
+    public int ResolvePageSize(int? requestPageSize) => ResolvePageSize(requestPageSize, DefaultPageSize, MaxPageSize);
```
PawelPawelec-RDX, line 110: I don't think all these resolve methods belong in configuration class, it should be in some separate class that takes request value and configuration limit.

PawelPawelec-RDX (changes requested):

src/RadixDlt.NetworkGateway.GatewayApi/Configuration/EndpointOptions.cs
krzlabrdx: I disagree. This is logic tightly bounded to the configuration itself. There's no point breaking the locality of the behavior for no good reason.

PawelPawelec-RDX (approved):

PR #750, opened 2024-05-23
Entity correlation

krzlabrdx:


src/RadixDlt.NetworkGateway.Abstractions/Model/EntityRelationship.cs
```diff
@@ -62,29 +62,29 @@
  * permissions under this License.
  */
 
-using Microsoft.EntityFrameworkCore.Migrations;
+namespace RadixDlt.NetworkGateway.Abstractions.Model;
 
-#nullable disable
-
-namespace RadixDlt.NetworkGateway.PostgresIntegration.Migrations
+// TODO change to enum
```
PawelPawelec-RDX, line 67: hmm I'm not sure, maintaining string might be easier. Also I'd make it more descriptive by nesting classes i.e
```
public static class EntityRelationship
{
    public static class Component
    {
        public const string Package
    }

    public static class Validator 
    {
    }

... etc
}
```

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/RelationshipProcessor.cs
```diff
@@ -0,0 +1,247 @@
...
+        }
+
+        if (substateData is CoreModel.TypeInfoModuleFieldTypeInfoSubstate typeInfoSubstate)
+        {
+            switch (typeInfoSubstate.Value.Details)
+            {
+                case CoreModel.ObjectTypeInfoDetails objectDetails:
+                    referencedEntity.PostResolveConfigure((ComponentEntity e) =>
+                    {
+                        e.AssignedModuleIds = objectDetails.ModuleVersions
```
PawelPawelec-RDX, line 103: it's not related to relationships. maybe we have to process same substate in multiple places to keep it clean.

src/RadixDlt.NetworkGateway.Abstractions/Model/EntityRelationship.cs
krzlabrdx: Had we decided to go with strings that would be no doubt the way to go. However I have rather strong objections here. We should go with enum as
- those relationships are well-defined and finite; we can always add new ones if needed,
- enums are lighter and easier to work with

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/RelationshipProcessor.cs
krzlabrdx: Ah, I noticed that myself and wanted to change that before pushing... obviously I must have forgotten about about that before I pushed.

PawelPawelec-RDX (approved):

PawelPawelec-RDX (approved):

PR #755, opened 2024-06-11
Resource holders endpoint.

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1898,6 +1898,31 @@ paths:
         "4XX":
           $ref: "#/components/responses/ErrorResponse"
 
+  /statistics/page/resource-owners/:
```
dhedey, line 1901: Possibly `/extensions/resource-owners/page` (in that it's a page under the `resource-owners` extension)?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -4748,6 +4830,23 @@ components:
             validators:
               $ref: "#/components/schemas/ValidatorUptimeCollection"
 
+    ResourceOwnersRequest:
+      allOf:
+      - $ref: "#/components/schemas/CursorLimitMixin"
```
dhedey, line 4835: 👍  Looks good for no ledger state mixin

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -4748,6 +4830,23 @@ components:
...
+      - type: object
+        properties:
+          resource_address:
```
dhedey, line 4838: I see we haven't got a Global / Parent / Vault aggregation option. I assume it's Globally aggregated?

dhedey (commented):
Nice 👍

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: we follow page/resource/ pattern in all other places, so i think we should stick with that.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: yes it's globally aggregated. do we need to vault aggregate as well? that would make it a bit more complex on database and aggregation side.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
dhedey: If we're using the `/extensions/` prefix, I'd like each extension to be conceptually separate (i.e. each extension `X` would serve routes under `/extensions/X/**`, which could let us have the extensions live elsewhere), and aligned with V2 design.

I'd rather that over aligning with the Gateway URI structure.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX, line 4838: @dhedey is it ok to query by just one resource? or should that endpoint support providing multiple resource addresses?

dhedey: Yeah, only one resource is fine. I think it only really makes sense to ask the question for each resource separately.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/WriteHelper.cs
```diff
@@ -735,6 +735,67 @@ public async Task<int> CopyKeyValueStoreSchemaHistory(ICollection<KeyValueStoreS
...
+                await writer.WriteNullAsync(token);
+                await writer.WriteAsync(nfe.TotalCount, NpgsqlDbType.Bigint, token);
+            }
+        }
+
+        await writer.CompleteAsync(token);
+        await writer.DisposeAsync();
+
+        await using var copyFromTempTablecommand = _connection.CreateCommand();
+        copyFromTempTablecommand.CommandText = @"
```
krzlabrdx, line 784: I hoped we'll not execute this as part of the regular DA flow as it may be really slow. How about:
1. Create long-lived temp table for those writes. i.e. just a regular table that the DA will append data to.
2. Move this logic to some db-level trigger.

This way DA won't need to wait for this operation to complete.

The query would remain essentially identical, maybe with an extra `WHERE old.from_state_version < new.from_state_version` check

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/ExtensionsQuerier.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        switch (resourceEntity)
+        {
+            case GlobalFungibleResourceEntity:
+            {
+                var totalCount = await _dbContext.ResourceOwners.CountAsync(x => x.ResourceEntityId == resourceEntity.Id, token);
+
+                var cd = new CommandDefinition(
+                    commandText: @"
+SELECT
```
krzlabrdx, line 113: We should use cursor-based pagination here. OFFSET will be too expensive.

krzlabrdx (changes requested):

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
```diff
@@ -1438,6 +1438,41 @@ void AggregateEntityResourceVaultInternal(long entityId, long resourceEntityId,
                 }
             }
 
+            var resourceOwnersToAdd = entityResourceAggregatedVaultsHistoryToAdd
```
krzlabrdx, line 1441: Actually the source collection is guaranteed to be ordered by `state_version` so we could simplify all of that to:
```
var resourceOwnersToAdd = new Dict(..);

foreach (var x in entityResourceAggregatedVaultsHistoryToAdd) {
  resourceOwnersToAdd.AddOrUpdate(
    (x.EntityId, x.ResourceEntityId), 
    _ => new NonFungibleResourceOwners(...), 
    existing => existing.Balance = ...)
}
```

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -70,6 +71,7 @@ tags:
       - **Stream** (`/transaction/*`) - For reading committed transactions.
       - **State** (`/state/*`) - For reading the current or past ledger state of the network.
       - **Statistics** (`/statistics/*`) - For calculating particular statistics against the current or past ledger state of the network.
+      - **Extensions** (`/Extensions/*`) - For additional endpoints that do not fit into other categories. They do not support historical browsing.
```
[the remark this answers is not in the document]
dhedey: Let's chat about this separately

CHANGELOG.md
```diff
@@ -18,13 +18,15 @@ Release built: _not released yet_
 - Added support for two-way linked dApps in the `/state/entity/details` endpoint.
   - Brand-new `two_way_linked_*` properties on the `details` property of Resources, Accounts, Packages and other global components.
   - See https://docs.radixdlt.com/docs/metadata-for-verification#metadata-standards-for-verification-of-onledger-entities for detailed specification.
+- Added new endpoint `/extensions/resource-owners/page` which returns information about all owners of the queried resource.
```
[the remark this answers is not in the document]
dhedey: Let's chat about this separately

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -4745,12 +4832,24 @@ components:
                 - kind: "U32"
                   value: "1"
     StateKeyValueStoreDataRequestKeyItem:
+#TODO PP: check if that's not a breaking change.
+      oneOf:
```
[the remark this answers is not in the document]
dhedey: I think this change is separate from the rest of the PR?

I think I'm in agreement with Krzysztof - this will break client generation:
* Different clients chose to deserialize `oneOf` differently
* In general we have avoided `oneOf` without discriminators because I think because this is inefficient with deserialization as they need to analyze / validate the payload to work out the type)

I would suggest it would be more useful to add comments to explain one or the other can be provided :). This is definitely one of those endpoints where more comments would be useful. But also happy to do this in a separate PR.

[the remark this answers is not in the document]
PawelPawelec-RDX: ok i'll extract that into separate PR, assumed that it will be straightforward so I did it here.

src/RadixDlt.NetworkGateway.GatewayApi/Services/IExtensionsQuerier.cs
```diff
@@ -0,0 +1,82 @@
...
+
+using RadixDlt.NetworkGateway.Abstractions;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace RadixDlt.NetworkGateway.GatewayApi.Services;
+
+public interface IExtensionsQuerier
+{
+    Task<GatewayApiSdk.Model.ResourceOwnersResponse> ResourceOwners(
```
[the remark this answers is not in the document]
PawelPawelec-RDX: done, fixed entire codebase. tbh if we want to be that strict with that using we should have some static analyzer for that

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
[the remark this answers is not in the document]
PawelPawelec-RDX: https://github.com/radixdlt/babylon-gateway/pull/780

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
```diff
@@ -1438,6 +1438,41 @@ void AggregateEntityResourceVaultInternal(long entityId, long resourceEntityId,
...
+                    (key, group) => new
+                    {
+                        key,
+                        newestEntry = group.MaxBy(x => x.FromStateVersion),
+                    }
+                )
+                .Select(
+                    x =>
+                    {
+                        ResourceOwners result = x.newestEntry switch
```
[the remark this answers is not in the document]
PawelPawelec-RDX: hmm ok, let's do that.

src/RadixDlt.NetworkGateway.GatewayApi/Services/IExtensionsQuerier.cs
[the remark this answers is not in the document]
krzlabrdx: Yes, we should. Same's true for the GatewayApi/Client, Core* and Toolkit*. Have you ever written such an analyzer?  The question is: is it worth it?

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
```diff
@@ -1451,17 +1451,50 @@ void AggregateEntityResourceVaultInternal(long entityId, long resourceEntityId,
                     _ => throw new ArgumentOutOfRangeException(nameof(x), x, null),
                 };
 
+                var isDeleted = balance.IsZero();
```
krzlabrdx, line 1454: How accurate the withdrawals / vault balances are? @dhedey ? I'm asking cause I wonder if it is reasonable to expect `0.000000000000001` - similarly to how you have to handle such cases with floating point.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
```diff
@@ -1523,7 +1556,7 @@ void AggregateEntityResourceVaultInternal(long entityId, long resourceEntityId,
             rowsInserted += await writeHelper.CopyResourceEntitySupplyHistory(resourceEntitySupplyHistoryToAdd, token);
             rowsInserted += await writeHelper.CopyNonFungibleDataSchemaHistory(nonFungibleSchemaHistoryToAdd, token);
             rowsInserted += await writeHelper.CopyKeyValueStoreSchemaHistory(keyValueStoreSchemaHistoryToAdd, token);
-            rowsInserted += await writeHelper.CopyResourceOwners(resourceOwnersToAdd.Select(x => x.Value).ToList(), token);
+            rowsInserted += await writeHelper.CopyResourceOwners(resourceOwnersToAdd, token);
```
krzlabrdx, line 1559: It seems way overcomplicated to me. I guess I'd suggest to:
- keep track of `0` balance as if it was a regular value - it will simplify entire codebase by quite some margin,
- on the database side use MERGE with INSERT / UPDATE or DELETE clause while merging the temp table with the target table

krzlabrdx: Think:
```sql
MERGE INTO resource_owners ro
USING tmp_resource_owners tmp
ON ro.entity_id = tmp.entity_id AND ro.resource_entity_id = tmp.resource_entity_id
WHEN MATCH AND ro.from_state_version < tmp.from_state_version AND tmp.balance = 0 THEN <<delete>>
WHEN MATCH AND ro.from_state_version < tmp.from_state_version THEN <<update>>
WHEN MATCH DO NOTHING
WHEN NOT MATCHED THEN <<insert>>
```

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
```diff
@@ -1438,7 +1438,7 @@ void AggregateEntityResourceVaultInternal(long entityId, long resourceEntityId,
                 }
             }
 
-            var resourceOwnersToAdd = new Dictionary<ResourceOwnersLookup, ResourceOwnersChange>();
+            var resourceOwnersToAdd = new Dictionary<(long EntityId, long ResourceEntityId), ResourceOwners>();
```
krzlabrdx, line 1441: Please define a lookup struct.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
```diff
@@ -1451,50 +1451,17 @@ void AggregateEntityResourceVaultInternal(long entityId, long resourceEntityId,
...
-
-                        existing.Entry = new ResourceOwners
-                        {
-                            Id = sequences.ResourceOwnersSequence++,
-                            EntityId = x.EntityId,
-                            ResourceEntityId = x.ResourceEntityId,
-                            Balance = balance,
-                        };
-                    });
+                    existing => existing.Balance = balance
```
krzlabrdx, line 1463: I think we should keep track of the `from_state_version`, shouldn't we?

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
```diff
@@ -1556,7 +1523,7 @@ void AggregateEntityResourceVaultInternal(long entityId, long resourceEntityId,
             rowsInserted += await writeHelper.CopyResourceEntitySupplyHistory(resourceEntitySupplyHistoryToAdd, token);
             rowsInserted += await writeHelper.CopyNonFungibleDataSchemaHistory(nonFungibleSchemaHistoryToAdd, token);
             rowsInserted += await writeHelper.CopyKeyValueStoreSchemaHistory(keyValueStoreSchemaHistoryToAdd, token);
-            rowsInserted += await writeHelper.CopyResourceOwners(resourceOwnersToAdd, token);
+            rowsInserted += await writeHelper.CopyResourceOwners(resourceOwnersToAdd.Select(x => x.Value).ToList(), token);
```
krzlabrdx, line 1526: `resourceOwnersToAdd.Values.ToList()`

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
PawelPawelec-RDX: what for? newest always wins

krzlabrdx: Ah, right now we execute MERGE synchronously so there's no need for the `from_state_version`. However didn't we want to include `last_updated_at_state_version` in the resource owners endpoint response?

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/ResourceOwnersProcessor.cs
```diff
@@ -0,0 +1,231 @@
...
+//
+// namespace RadixDlt.NetworkGateway.PostgresIntegration.LedgerExtension;
+//
+// public record ResourceOwner(EntityAddress OwnerAddress);
+//
+// public record FungibleResourceOwner(EntityAddress OwnerAddress, string Amount) : ResourceOwner(OwnerAddress);
+//
+// public record NonFungibleResourceOwner(EntityAddress OwnerAddress, long NfidCount) : ResourceOwner(OwnerAddress);
+//
+// internal class ResourceOwnersProcessor
```
krzlabrdx, line 83: Drop this file.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/ExtensionsQuerier.cs
```diff
@@ -0,0 +1,177 @@
...
+INNER JOIN entities e
+ON ro.entity_id = e.id
+WHERE ro.resource_entity_id = @resourceEntityId
+  AND (@balanceBoundary is null OR ((ro.balance, ro.id) < (Cast(@balanceBoundary AS numeric(1000,0)), @idBoundary)))
+ORDER BY (ro.balance, ro.entity_id) DESC
+LIMIT @limit",
+            parameters: new
+            {
+                resourceEntityId = resourceEntity.Id,
+                balanceBoundary = cursor?.BalanceBoundary,
```
krzlabrdx, line 132: Guess it would be easier to default balance boundary to `0`? No extra if-o-logy in the query itself.

krzlabrdx: Ah, we'd have to default it to "max". I think `TokenAmount` should actually define such a construct.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/ExtensionsQuerier.cs
```diff
@@ -0,0 +1,177 @@
...
+                idBoundary = cursor?.IdBoundary ?? long.MaxValue,
+                limit = limit,
```
krzlabrdx, line 134: Shouldn't we use `limit + 1`?

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
PawelPawelec-RDX: @dhedey do we want to include that? that wasn't part of initially approved specs.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
PawelPawelec-RDX: it's no longer relevant as I've removed that piece of code.

src/RadixDlt.NetworkGateway.GatewayApi/Services/IExtensionsQuerier.cs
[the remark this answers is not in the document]
PawelPawelec-RDX: I don't think it's worth it, if we want to use that in entire codebase maybe it's better to have global usings.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/ExtensionsQuerier.cs
krzlabrdx: It seems that scrypto defines `Decimal.MAX` as `I192::MAX`. This could be encoded in our `TokenAmount`. Should have no impact on performance and cleaner code is of higher priority if you ask me.

krzlabrdx (approved):

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/WriteHelper.cs
PawelPawelec-RDX: After discussion we decided to go with current solution and check performance on some real network.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs
dhedey: Hm this is a good question.

Yes, I think in retrospect, thinking through the kind of insights I'd want, it would be very useful to store `last_updated_at_state_version` as an additional field (from state version is the wrong naming / abstraction).

And it would be useful to include that on the response too, if it'd be straightforward.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1900,6 +1905,33 @@ paths:
...
           $ref: "#/components/responses/ErrorResponse"
 
+  /extensions/resource-owners/page:
+    post:
+      operationId: ResourceOwnersPage
+      summary: Get Resource Owners Page
+      description: |
+        Returns list of all owners of a given resource, ordered by Amount (fungibles)/ Number of items (non fungibles) descending.
+        This endpoint operates only at the current state version, it is not possible to browse historical data.
+        Because of that, it is not possible to offer stable pagination as data constantly changes. Values might change, which might result in gaps or some entries being returned twice.
```
dhedey, line 1915: I realise I never properly discussed what this endpoint was called on the ticket.

I'm sorry if this feels like a pivot, but can we rename this? As per the name of NG-447, I'd like to call this "Resource Holders" not "Resource Owners", to avoid confusion with the _owner_ of the resources in the engine auth role sense.

To that end, can I ask we rename the various files/names to align? (I guess we can do a CTRL-F for "owner" and fix them in turn).

As for these descriptions:

```suggestion
  /extensions/resource-holders/page:
    post:
      operationId: ResourceHoldersPage
      summary: Resource Holders
      description: |
        A paginated endpoint to discover which global entities hold the most of a given resource.

        More specifically, it returns a page of global entities which hold the given resource, ordered descending by the total fungible balance / total count of non-fungibles stored in vaults in the state tree of that entity (excluding unclaimed royalty balances).

        This endpoint operates only at the **current state version**, it is not possible to browse historical data.

        Because of that, it is not possible to offer stable pagination as data constantly changes. Balances might change between pages being read, which might result in gaps or some entries being returned twice.

        The default maximum page size is X.
```

Can we also spell out what `X` is? As we work through these endpoints, we need to record X so people can work out what their max page size requested is, without getting a 400.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1195,6 +1197,9 @@ tags:
   - name: Statistics
     x-displayName: Statistics Endpoints
     description: Calculate particular statistics based on current or historic ledger state.
+  - name: Extensions
+    x-displayName: Extensions Endpoints
+    description: Additional endpoints that do not fit into other categories.
```
dhedey, line 1202:
```suggestion
    description: Endpoints that do not fit into other categories, for example, those which provide custom aggregation or indexing, or don't support historic state.
```

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
dhedey, line 74:
```suggestion
      - **Extensions** (`/Extensions/*`) - For endpoints that do not fit into other categories, for example, they may not support historic state.
```

src/RadixDlt.NetworkGateway.GatewayApi/Handlers/DefaultExtensionsHandler.cs
```diff
@@ -0,0 +1,87 @@
...
+
+internal class DefaultExtensionsHandler(IExtensionsQuerier extensionsQuerier, IOptionsSnapshot<EndpointOptions> endpointConfiguration) : IExtensionsHandler
+{
+    public async Task<GatewayModel.ResourceOwnersResponse> ResourceOwners(GatewayModel.ResourceOwnersRequest request, CancellationToken token)
+    {
+        var cursor = GatewayModel.ResourceOwnersCursor.FromCursorString(request.Cursor);
+
+        return await extensionsQuerier.ResourceOwners(
+            (EntityAddress)request.ResourceAddress,
+            endpointConfiguration.Value.ResolvePageSize(request.LimitPerPage),
```
dhedey, line 83: What is the maximum page size realistically?

Maybe we can have a default of 100, max of 1000? If it wouldn't be too expensive to query and return?

dhedey (changes requested):
Really nice!

A couple of key comments - a slightly annoying reflection on naming; and a question around page size.

src/RadixDlt.NetworkGateway.GatewayApi/Handlers/DefaultExtensionsHandler.cs
PawelPawelec-RDX: currently default and max are same and set to 100. Why would anyone need to fetch 1000 with one call? that sounds unreasonable.

krzlabrdx (approved):

src/RadixDlt.NetworkGateway.GatewayApi/Handlers/DefaultExtensionsHandler.cs
PawelPawelec-RDX: ok, will put 100 as default and 1000 as max.

dhedey (approved):

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -1900,6 +1905,35 @@ paths:
...
+  /extensions/resource-holders/page:
+    post:
+      operationId: ResourceHoldersPage
+      summary: Get Resource Holders Page
+      description: |
+        A paginated endpoint to discover which global entities hold the most of a given resource.
+        More specifically, it returns a page of global entities which hold the given resource, ordered descending by the total fungible balance / total count of non-fungibles stored in vaults in the state tree of that entity (excluding unclaimed royalty balances).
+        This endpoint operates only at the **current state version**, it is not possible to browse historical data.
+        Because of that, it is not possible to offer stable pagination as data constantly changes. Balances might change between pages being read, which might result in gaps or some entries being returned twice.
+        The default maximum page size is 1000.
```
dhedey, line 1917: I think this is a little confusing - people will expect the default page size to be 100. Perhaps the following is a little clearer?

```suggestion
Under default Gateway configuration, up to 100 entries are returned per response. This can be increased up to 1000 entries per page with the `limit_per_page` parameter.
``````

dhedey: Have committed this for you to save you the hassle :)

PR #780, opened 2024-08-09
describe that it's required to provide only one format of key.

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.GatewayApi/Validators/StateKeyValueStoreDataRequestKeyItemValidator.cs
```diff
@@ -76,5 +76,8 @@ public StateKeyValueStoreDataRequestKeyItemValidator()
 
         RuleFor(x => x)
             .Must(x => x.KeyHex != null || x.KeyJson != null).WithMessage("Either key_hex or key_json must be not empty.");
+
+        RuleFor(x => x)
```
PawelPawelec-RDX, line 80: That will be a breaking change so i'm not 100% we want to do that, but providing them both at once will lead into non deterministic results so I think it's better to fail early.

krzlabrdx: There's literally the exact same rule literally a line above. This is already covered. :D

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -4745,6 +4745,7 @@ components:
                 - kind: "U32"
                   value: "1"
     StateKeyValueStoreDataRequestKeyItem:
+      description: It is required to provide either key_hex or key_json property.
```
krzlabrdx, line 4748: Please wrp property names in backticks to improve formatting.

src/RadixDlt.NetworkGateway.GatewayApi/Validators/StateKeyValueStoreDataRequestKeyItemValidator.cs
PawelPawelec-RDX: > What's more our current code behaves in a deterministic manner: if key_hex is available then we use it, otherwise we fallback to key_json.

I'm not sure we should do that:P

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
dhedey: It sounds a little clunky. Just the following is fine I think:
```suggestion
      description: Provide either `key_json` or `key_hex`.
```
.

src/RadixDlt.NetworkGateway.GatewayApi/Validators/StateKeyValueStoreDataRequestKeyItemValidator.cs
dhedey: Sadly (based on seeing how the radquest team initially misunderstood this endpoint and initially passed both), I think it's best to NOT add this validation and risk breaking clients.

Instead, we should just document it in the docs:
`Provide either `key_json` or `key_hex`. If both are provided, `key_hex` is used and `key_json` is ignored.`

dhedey (approved):

PR #782, opened 2024-08-14
Entity resource + Vault aggregate rework (DA)

krzlabrdx:


src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/EntityResourceProcessor.cs
```diff
@@ -0,0 +1,483 @@
...
+    public async Task LoadDependencies()
+    {
+        _existingResourceDefinitions.AddRange(await ExistingEntityResourceEntryDefinitions());
+        _mostRecentResourceTotalsHistory.AddRange(await MostRecentEntityResourceTotalsHistory());
+        _mostRecentResourceBalanceHistory.AddRange(await MostRecentResourceBalanceHistory());
+        _existingResourceVaultDefinitions.AddRange(await ExistingEntityResourceVaultEntryDefinitions());
+        _mostRecentResourceVaultTotalsHistory.AddRange(await MostRecentEntityResourceVaultTotalsHistory());
+    }
+
+    public void ProcessChanges()
```
PawelPawelec-RDX, line 157: Do we need to process all these changes in one "processor"? Is it possible to extract smaller ones? Or at least divide that logic into separate methods?

krzlabrdx: Yeah, it seems to be the case. I guess those foreach loop bodies could have been split into separate method though.

PR #791, opened 2024-09-04
1.7.2 -> main

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.GatewayApi/Services/TransactionPreviewService.cs
```diff
@@ -165,7 +165,8 @@ public TransactionPreviewService(INetworkConfigurationProvider networkConfigurat
             nonce: request.Nonce,
             signerPublicKeys: request.SignerPublicKeys.Select(MapPublicKey).ToList(),
             message: message,
-            flags: coreRequestFlags);
+            flags: coreRequestFlags,
+            options: new CoreModel.TransactionPreviewResponseOptions(request.OptIns?.RadixEngineToolkitReceipt ?? false));
```
dhedey, line 169: Hmm I wonder if an older Core API accepts this field.

If it errors, perhaps to improve compatibility and ensure this works with an existing bottlenose node, we could set options to `null` (which I believe would map to not adding the options property) unless it's true?

```
    options = (request.OptIns?.RadixEngineToolkitReceipt ?? false) ? new CoreModel.TransactionPreviewResponseOptions(true) : null
```

dhedey (approved):

PR #796, opened 2024-09-25
Configuration to reduce database size

PawelPawelec-RDX:


CHANGELOG.md
```diff
@@ -27,12 +27,20 @@ Release built: _not released yet_
 >   - `/state/entity/page/non-fungibles/` (when using `non_fungible_include_nfids` opt-in)
 >   - `/state/entity/page/non-fungible-vaults/` (when using `non_fungible_include_nfids` opt-in)
 
+### What’s new?
+- New configuration options `DataAggregator__Storage__StoreTransactionReceiptEvents`, and `DataAggregator__Storage__StoreReceiptStateUpdates` for data aggregator to configure if transaction's receipt events and receipt state updates should be stored in the database. It is meant to be used by gateway runners who want to reduce their database size. Keep in mind that when disabled `/stream/transactions` and `/stream/transactions` will not return that data.
```
dhedey, line 31: A few things:
* `/stream/transactions` is repeated twice - do you mean a different endpoint?
* On first reading, it sounds like `/stream/transactions` won't work? Do you mean the receipt fields for the given transactions return `null` or `{}` or is the field just missing? or something? 

I think we want:
```suggestion
- New configuration options `DataAggregator__Storage__StoreTransactionReceiptEvents`, and `DataAggregator__Storage__StoreReceiptStateUpdates` for the data aggregator to configure if a transaction's receipt events and receipt state updates should be stored in the database. It is meant to be used by gateway runners who want to reduce their database size. Keep in mind that when disabled, the corresponding properties will be missing on the `/stream/transactions` response for filtered out transactions. You can save significant space by using `StoryOnlyForUserTransactionsAndEpochChanges` and only excluding round change transactions, which aren't typically read from the `/stream/transactions` endpoint.
```

CHANGELOG.md
```diff
@@ -59,6 +67,8 @@ Release built: _not released yet_
         - Renamed `entity_vault_history` to `vault_balance_history`. Holds information about vault content (amount of fungibles or count of non fungible ids inside vault) at a given state version.
     - Key value store
         - New `key_value_store_totals_history` table, which holds total count of all keys under a given store at a given state version.
+- Changed `receipt_state_updates` in the `ledger_transactions` table to be nullable.
+- Moved all `receipt_event_*` columns from `ledger_transactions` to separate `ledger_transaction_events` table.
```
dhedey, line 71: Out of interest, why did we move the events to a separate table, but not also the state updates?

docs/configuration.md
```diff
@@ -73,6 +73,10 @@ The Network Gateway services can be configured in line with the [configuration i
 - `DataAggregator__Network__MaxAllowedStateVersionLagToBeConsideredSynced` (type: `int`, default value: `100`) - maximum allowed state version lag for CoreAPI node to be considered synced.
 - `DataAggregator__Network__IgnoreNonSyncedNodes` (type: `bool`, default value: `true`) - controls if nodes with a different status than healthy and synced should be used by the Data Aggregator.
 
+#### Storage
+- `DataAggregator__Storage__StoreTransactionReceiptEvents` (type: `enum`, default value: `StoreForAllTransactions`) - controls if data aggregator should store transaction receipt events in database.
```
dhedey, line 77: We should list out all supported values here, otherwise they're not discoverable enough.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -2317,16 +2317,14 @@ components:
...
-        - $ref: "#/components/schemas/ResultSetCursorMixin"
-        - type: object
-          required:
-            - items
-          properties:
-            items:
-              type: array
-              items:
-                $ref: "#/components/schemas/TwoWayLinkedDappsCollectionItem"
+      type: object
```
dhedey, line 2320: I'd have thought removing `ResultSetCursorMixin` is a breaking schema change?

I think we may be forced to keep it sadly, else people using a modern SDK might be broken when trying to read these endpoints after its removal.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -2337,16 +2335,14 @@ components:
 
     TwoWayLinkedEntitiesCollection:
       description: A collection of two-way linked (resolved & verified) entities linked to the entity defining this collection.
-      allOf:
```
dhedey, line 2340: Ditto here?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -4055,11 +4051,14 @@ components:
     StreamTransactionsResponse:
       allOf:
         - $ref: "#/components/schemas/LedgerStateMixin"
-        - $ref: "#/components/schemas/ResultSetCursorMixin"
         - type: object
           required:
             - items
           properties:
+            next_cursor:
```
dhedey, line 4058: I take it we never return a `total_count`? So makes sense to not use the `ResultSetCursorMixin`, I agree

dhedey (approved):
The core of this looks good, happy for you to merge once markups done 👍

CHANGELOG.md
PawelPawelec-RDX: because all events_ columns related to events and they are logically separate concept to rest of receipt. I don't think it makes sense to create separate table for just one column. i.e we could in future create one for all "raw" receipt data that's basically forwarded from core API.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: yes it was never returned here and was optional.

src/RadixDlt.NetworkGateway.PostgresIntegration/CommonDbContext.cs
```diff
@@ -81,15 +81,10 @@ internal abstract class CommonDbContext : DbContext
 {
     internal const string DiscriminatorColumnName = "discriminator";
 
-    /// <summary>
-    /// Gets LedgerTransactions.
-    /// </summary>
-    /// <remarks>
-    /// A LedgerTransaction row contains large blobs, so you must SELECT the fields you need after using this, and not pull down the whole
```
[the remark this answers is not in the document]
PawelPawelec-RDX: Yes it's true but I had a feeling that's unnecessary and provides no value, if you feel strong about it I can revert it.

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
PawelPawelec-RDX: yes they were optional and we've never returned them.. there's no endpoint to paginate over that :D

CHANGELOG.md
PawelPawelec-RDX: ok committed slightly altered suggestion (it also affects committed-details endpoint).

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/TransactionMapper.cs
```diff
@@ -0,0 +1,146 @@
...
+        {
+            ErrorMessage = lt.ReceiptErrorMessage,
+            Status = MapTransactionStatus(lt.ReceiptStatus),
+            Output = optIns.ReceiptOutput && lt.ReceiptOutput != null ? new JRaw(lt.ReceiptOutput) : null,
+            FeeSummary = optIns.ReceiptFeeSummary ? new JRaw(lt.ReceiptFeeSummary) : null,
+            FeeDestination = optIns.ReceiptFeeDestination && lt.ReceiptFeeDestination != null ? new JRaw(lt.ReceiptFeeDestination) : null,
+            FeeSource = optIns.ReceiptFeeSource && lt.ReceiptFeeSource != null ? new JRaw(lt.ReceiptFeeSource) : null,
+            CostingParameters = optIns.ReceiptCostingParameters ? new JRaw(lt.ReceiptCostingParameters) : null,
+            NextEpoch = lt.ReceiptNextEpoch != null ? new JRaw(lt.ReceiptNextEpoch) : null,
+            StateUpdates = optIns.ReceiptStateChanges ? new JRaw(lt.ReceiptStateUpdates) : null,
```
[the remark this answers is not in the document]
PawelPawelec-RDX: i'll change that so it's more readable but i did test that and property wasn't included.

PR #799, opened 2024-10-08
Do not index manifest for failed transactions

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/ManifestProcessor.cs
```diff
@@ -108,128 +108,169 @@ public void OnTransactionScan(CoreModel.CommittedTransaction transaction, long s
                 _referencedEntities.MarkSeenAddress(address);
             }
 
-            _manifestExtractedAddresses.Add(stateVersion, extractedAddresses);
+            if (transaction.Receipt.Status == CoreModel.TransactionStatus.Succeeded)
+            {
+                _manifestExtractedAddresses.Add(stateVersion, extractedAddresses);
```
dhedey, line 113: Out of interest, why do we only include this, and not e.g. `_referencedEntities` above? What is the `_referencedEntities` check used for?

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/ManifestProcessor.cs
```diff
@@ -108,128 +108,169 @@ public void OnTransactionScan(CoreModel.CommittedTransaction transaction, long s
...
                         new ManifestAddressLedgerTransactionMarker
                         {
                             Id = _context.Sequences.LedgerTransactionMarkerSequence++,
                             StateVersion = stateVersion,
                             OperationType = LedgerTransactionMarkerOperationType.AccountOwnerMethodCall,
-                            EntityId = re.DatabaseId,
+                            EntityId = referencedEntity.DatabaseId,
                         });
                 }
+                else if (addressWithEntityType.EntityType is not (ToolkitModel.EntityType.GlobalVirtualEd25519Account or ToolkitModel.EntityType.GlobalVirtualSecp256k1Account))
```
dhedey, line 229: This is incorrect and flakey (e.g. it misses virtual identities).

Could we instead have a `is_preallocated()` method on the address type, which delegates to an `is_preallocated()` method on the entity type?

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/ManifestProcessor.cs
```diff
@@ -108,128 +108,169 @@ public void OnTransactionScan(CoreModel.CommittedTransaction transaction, long s
...
+                        OperationType = LedgerTransactionMarkerOperationType.ResourceInUse,
+                        EntityId = referencedEntity.DatabaseId,
+                    });
             }
 
-            foreach (var address in extractedAddresses.AccountsRequiringAuth)
+            foreach (var addressWithEntityType in extractedAddresses.AccountsRequiringAuth)
             {
-                if (_referencedEntities.TryGet(address, out var re))
+                if (_referencedEntities.TryGet(addressWithEntityType.Address, out var referencedEntity))
```
dhedey, line 218: I'm a bit confused. It looks like `_referencedEntities` just extracts all addresses from the manifest, but at this stage, we haven't checked that these addresses either exist on ledger already, right? (or were created in this transaction?)

Or does this happen _after_ the resolution of the `_referencedEntities`? i.e. at this point, we know that all `_referencedEntities` either exist on ledger already, or were created in this transaction (or previous transaction in the batch)? What happens at resolution time if `_referencedEntities` don't exist (and aren't preallocated addresses)? I assume that we already throw?

Either way, I don't think the check here really makes sense? Perhaps we should discuss on a call?

dhedey: Following conversations with Pawel:
* At this point, we have done a DB load of the `MarkSeenAddress` addresses
* We _don't throw_ if an address doesn't match - we just don't see it in the `TryGet` method.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/ManifestProcessor.cs
dhedey: As discussed, we can probably move this if statement to cover `_referencedEntities.MarkSeenAddress(address);` as well, but we need to check for any edge cases.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/ManifestProcessor.cs
dhedey: So this looks good 👍

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/ManifestProcessor.cs
PawelPawelec-RDX: moved.  We still need to extract manfiest to do classification as we do that for failed tx as well.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/ManifestAddressesExtractor.cs
```diff
@@ -114,22 +114,36 @@ public static ManifestAddresses ExtractAddresses(ToolkitModel.TransactionManifes
...
         var resourceAddresses = allAddresses
             .Where(x => x.Key is ToolkitModel.EntityType.GlobalFungibleResourceManager or ToolkitModel.EntityType.GlobalNonFungibleResourceManager)
             .SelectMany(x => x.Value.Select(y => (EntityAddress)y.AddressString()))
             .ToList();
+
         var accountAddresses = allAddresses
-            .Where(x => x.Key is ToolkitModel.EntityType.GlobalAccount or ToolkitModel.EntityType.GlobalVirtualEd25519Account
-                or ToolkitModel.EntityType.GlobalVirtualSecp256k1Account)
+            .Where(
+                x => x.Key
```
dhedey, line 134: OPTIONAL / MINOR: By the same logic, ideally we'd add helper extension methods `IsPackage()` / `IsGenericComponent` / `IsResource` / `IsAccount` / `IsIdentity` etc on `ToolkitModel.EntityType`

No need to do it in this PR though

dhedey (approved):
LGTM, thanks 👍

PR #801, opened 2024-10-11
Fix missing round updates for stream transactions

PawelPawelec-RDX:


CHANGELOG.md
```diff
@@ -72,7 +73,8 @@ Release built: _not released yet_
     - Key value store
         - New `key_value_store_totals_history` table, which holds total count of all keys under a given store at a given state version.
 - Changed `receipt_state_updates` in the `ledger_transactions` table to be nullable.
-- Moved all `receipt_event_*` columns from `ledger_transactions` to separate `ledger_transaction_events` table.
+- Moved all `receipt_event_*` columns from the `ledger_transactions` to the separate `ledger_transaction_events` table.
```
dhedey, line 76: MINOR: I'd probably phrase this as:
```suggestion
- Moved all `receipt_event_*` columns from the `ledger_transactions` table to a new separate `ledger_transaction_events` table.
```

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/OriginLedgerTransactionMarkerProcessor.cs
```diff
@@ -91,16 +96,22 @@ public void VisitTransaction(CoreApiSdk.Model.CommittedTransaction committedTran
...
-                {
-                    Id = _context.Sequences.LedgerTransactionMarkerSequence++,
-                    StateVersion = stateVersion,
-                    OriginType = LedgerTransactionMarkerOriginType.User,
-                });
-        }
+            FlashLedgerTransaction => LedgerTransactionMarkerOriginType.Flash,
+            GenesisLedgerTransaction => LedgerTransactionMarkerOriginType.Genesis,
+            RoundUpdateLedgerTransaction => LedgerTransactionMarkerOriginType.RoundUpdate,
+            UserLedgerTransaction => LedgerTransactionMarkerOriginType.User,
```
dhedey, line 104: This is fine as-is, but conceptually, we can split a `LedgerTransaction` origin as `User`, `Validator` (round update) and `System` (genesis or protocol update = flash). I think this split is perhaps a little more instructive, but I don't feel super strongly if you feel this is much more consistent. Interested to hear your thoughts.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/OriginLedgerTransactionMarkerProcessor.cs
```diff
@@ -91,16 +96,22 @@ public void VisitTransaction(CoreApiSdk.Model.CommittedTransaction committedTran
...
+            _ => throw new ArgumentOutOfRangeException($"Unexpected ledger transaction type: {committedTransaction.LedgerTransaction.GetType()}"),
+        };
+
+        _ledgerTransactionMarkersToAdd.Add(
+            new OriginLedgerTransactionMarker
```
dhedey, line 109: I take it these markers aren't actually exposed? I'm slightly surprised we don't need to update the Gateway API? Or does it just match on `LedgerTransactionMarkerOriginType.User`?

Out of interest, how does it distinguish epoch changes from round updates?

dhedey (approved):
Approving, with a couple of questions around the origin type.

Also, just to check, we're adding some tests for the "All" kind filter?

PR #816, opened 2024-12-05
Hotfix/v1.9.2

PawelPawelec-RDX:


CHANGELOG.md
```diff
@@ -79,7 +90,7 @@ Release built: 23.10.2024
 - New configuration options `DataAggregator__Storage__StoreTransactionReceiptEvents`, and `DataAggregator__Storage__StoreReceiptStateUpdates` for the data aggregator to configure if a transaction's receipt events and receipt state updates should be stored in the database. It is meant to be used by gateway runners who want to reduce their database size. Keep in mind that when disabled, the corresponding properties will be missing on a response from both the `/stream/transactions` and the `/transaction/committed-details` endpoints. You can save significant space by using `StoryOnlyForUserTransactionsAndEpochChanges` and only excluding round change transactions, which aren't typically read from the `/stream/transactions` endpoint.
   - Possible values:
     - `StoreForAllTransactions` (default) - will store data for all transactions.
-    - `StoryOnlyForUserTransactionsAndEpochChanges` - will store data for user transactions and transactions that resulted in epoch change.
+    - `StoreOnlyForUserTransactionsAndEpochChanges` - will store data for user transactions and transactions that resulted in epoch change.
```
dhedey, line 93: Well... to be accurate, this should probably stay as `StoryOnlyForUserTransactionsAndEpochChanges` in the Changelog, with a highlight that the typo is fixed in 1.9.2 :D 

Indeed, I've already added a comment below which should be amended to say 1.9.2 instead of "future version"

CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
-## 1.10.0
+## 1.9.2
 Release built: _not released yet_
 
+### What’s new?
+- Added support for pre-allocated, non-persisted accounts in the `/state/account/page/resource-preferences` and `/state/account/page/authorized-depositors` endpoints.
```
dhedey, line 5: This is a bug-fix really rather than a feature addition.

CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
...
+
+### Bug fixes
+- Fixed a typo in the value `StoryOnlyForUserTransactionsAndEpochChanges` (replacing `Story` with `Store`) for the configuration entries `DataAggregator__Storage__StoreTransactionReceiptEvents` and `DataAggregator__Storage__StoreReceiptStateUpdates`. It now supports both values:
+  - `StoreOnlyForUserTransactionsAndEpochChanges`
+  - `StoryOnlyForUserTransactionsAndEpochChanges`
+- Fixed `/state/entity/details` when querying with `native_resource_details: true` option enabled for
```
dhedey, line 11: This should probably be top of the list as it's most important. Perhaps we should also be explicit that this is the 500 people were seeing, e.g.:

```suggestion
- Fixes a HTTP 500 response issue when querying the `/state/entity/details` endpoint with the `native_resource_details: true` opt-in for:
```

dhedey (approved):

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -2593,20 +2592,42 @@ components:
...
-          type: object
+          $ref: "#/components/schemas/ComponentEntityRoleAssignmentEntryAssignment"
+        updater_roles:
+          type: array
+          items:
+            $ref: "#/components/schemas/RoleKey"
+
+    ComponentEntityRoleAssignmentEntryAssignment:
+      type: object
+      required:
```
dhedey, line 2603: MINOR: Whilst we're here, is it easy to add a description?

Resolving the rule for this role:
* If it says `Explicit`, the rule can be read off the `explicit_rule` property on this object.
* If it's `Owner`, the rule can be read off `role_assignments.owner.rule` off the role assignments object two levels up.

dhedey (approved):

CHANGELOG.md
PawelPawelec-RDX: you're right, if someone would use 1.8.x he has to use value with typo :D reverted.

PR #817, opened 2024-12-12
Tx stream resource balance change and transaction status filter

PawelPawelec-RDX:


CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+## 1.10.0
+Release built: _not released yet_
+
+### API Changes
+- Two new filters to `/stream/transactions` endpoint.
+  - `transaction_status_filter` - Allows filtering by transaction status (`Successful`,`Failed`, `All`). Defaults to `All`.
+  - `balance_change_resources_filter` - Allows specifying an array of resources. If specified, the response will contain transactions which included non-fee related balance changes for all resources.
```
dhedey, line 7:
```suggestion
- New filters are supported on the `/stream/transactions` endpoint:
  - `transaction_status_filter` - Allows filtering by the transaction commit status (`Success`,`Failure`, `All`). Defaults to `All`.
  - `balance_change_resources_filter` - Allows filtering to transactions which included non-fee related balance changes for all provided resources. Defaults to `[]`. We recommend integrators use this instead of the `manifest_resources_filter` in most instances.

File: src/RadixDlt.NetworkGateway.Abstractions/Model/LedgerTransactionStatusFilter.cs [hunk 2]
```diff
@@ -0,0 +1,72 @@
...
+ * or non-commercial purpose and for any reproduction or redistribution by You of the
+ * Work. You assume all risks associated with Your use of the Work and the exercise of
+ * permissions under this License.
+ */
+
+namespace RadixDlt.NetworkGateway.Abstractions.Model;
+
+public enum LedgerTransactionStatusFilter
+{
+    Successful,
```
dhedey, line 69: Out of interest, did we consider `Success` and `Failure` instead?

The Core API uses `Success` and `Failure`, as does the Gateway API transaction status endpoint: https://radix-babylon-gateway-api.redoc.ly/#operation/TransactionStatus - perhaps we could consider changing?

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -4357,6 +4357,19 @@ components:
...
+              items:
+                $ref: "#/components/schemas/Address"
+            transaction_status_filter:
+              description: Allows filtering by transaction status. Defaults to `All`.
+              type: string
+              default: All
+              enum:
+                - Successful
+                - Failed
+                - All
```
dhedey, line 4372: If you agree with the change, perhaps:

```suggestion
                - Success
                - Failure
                - All
```

src/RadixDlt.NetworkGateway.GatewayApi/gateway-api-schema.yaml
```diff
@@ -4357,6 +4357,19 @@ components:
               type: array
               items:
                 $ref: "#/components/schemas/Address"
+            balance_change_resources_filter:
```
dhedey, line 4360: Can we also tweak a few other places:
* Could we add a description of these filters to the `Supported filters` section of the examples?
* Could we update the `manifest_resources_filter` to recommend that integrators likely wish to use `balance_change_resources_filter` instead in most cases, as it also captures resource changes which aren't in the manifest.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/ResourceBalanceChangeProcessor.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        _context = context;
+        _referencedEntities = referencedEntities;
+    }
+
+    public void VisitTransaction(CommittedTransaction transaction, long stateVersion)
+    {
+        var fungibleResourceBalanceChanges = transaction.BalanceChanges.FungibleEntityBalanceChanges
+            .SelectMany(x => x.NonFeeBalanceChanges)
+            .Select(x => _referencedEntities.Get((EntityAddress)x.ResourceAddress).DatabaseId)
```
dhedey, line 90: I assume we're guaranteed to have loaded these as referenced entities via handling the vault state changes or something? Maybe worth a comment if it's not immediately obvious to you?

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/TransactionTypeLedgerTransactionMarkerProcessor.cs
```diff
@@ -81,13 +81,17 @@ public TransactionTypeLedgerTransactionMarkerProcessor(ProcessorContext context)
 
     public void VisitTransaction(CoreModel.CommittedTransaction committedTransaction, long stateVersion)
     {
+        if (committedTransaction.LedgerTransaction is CoreModel.RoundUpdateLedgerTransaction)
```
dhedey, line 84: VERY MINOR: If it's possible in C Sharp, maybe it would be slightly more clear that we've covered all cases if we inline the return into the match as:
```csharp
CoreModel.RoundUpdateLedgerTransaction => {
    // We purposefully don't tag round updates, because we don't allow filtering on them.
    // And as there are quite a few round updates, this improves ingestion and querying performance.
    return;
}
```

(Also with a comment explaining why)

src/RadixDlt.NetworkGateway.PostgresIntegration/Migrations/IdempotentApplyMigrations.sql
```diff
@@ -1450,3 +1450,22 @@ BEGIN
...
+    IF NOT EXISTS(SELECT 1 FROM "__EFMigrationsHistory" WHERE "MigrationId" = '20241211105351_NewResourceBalanceChangeTransactionMarker') THEN
+    ALTER TYPE ledger_transaction_marker_type ADD VALUE 'resource_balance_change';
+    END IF;
+END $EF$;
+
+DO $EF$
+BEGIN
+    IF NOT EXISTS(SELECT 1 FROM "__EFMigrationsHistory" WHERE "MigrationId" = '20241211105351_NewResourceBalanceChangeTransactionMarker') THEN
+    INSERT INTO "__EFMigrationsHistory" ("MigrationId", "ProductVersion")
+    VALUES ('20241211105351_NewResourceBalanceChangeTransactionMarker', '8.0.2');
```
dhedey, line 1467: This doesn't include the removal of the round update transaction markers - perhaps it needs regenerating?

src/RadixDlt.NetworkGateway.PostgresIntegration/Models/LedgerTransactions/LedgerTransactionMarker.cs
```diff
@@ -72,7 +72,7 @@ namespace RadixDlt.NetworkGateway.PostgresIntegration.Models;
 internal enum LedgerTransactionMarkerTransactionType
 {
     User,
-    RoundChange,
+    RoundChange, // TODO PP: It can be removed safely after next resync. Without resync we should keep it here even though it's not used.
```
dhedey, line 75: :+1:

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/TransactionQuerier.cs
```diff
@@ -182,10 +183,17 @@ public async Task<TransactionPageWithoutTotal> GetTransactionStream(TransactionS
             ? request.Cursor?.StateVersionBoundary ?? request.FromStateVersion
             : request.FromStateVersion;
 
-        var searchQuery = _dbContext
-            .LedgerTransactionMarkers
-            .Where(lt => lt.StateVersion <= upperStateVersion && lt.StateVersion >= (lowerStateVersion ?? lt.StateVersion))
-            .Select(lt => lt.StateVersion);
+        var baseQuery = _dbContext
+            .LedgerTransactions
+            .Where(lt => lt.StateVersion <= upperStateVersion && lt.StateVersion >= (lowerStateVersion ?? lt.StateVersion));
```
dhedey, line 188: Out of interest, have we tested this perf against a prod database?
I assume it will be a bit faster, as it just uses the PK index; but might be worth verifying

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/TransactionQuerier.cs
```diff
@@ -182,10 +183,17 @@ public async Task<TransactionPageWithoutTotal> GetTransactionStream(TransactionS
...
+
+        var searchQuery = request.SearchCriteria.Status switch
+        {
+            LedgerTransactionStatusFilter.All => baseQuery.Select(lt => lt.StateVersion),
+            LedgerTransactionStatusFilter.Successful => baseQuery.Where(x => x.ReceiptStatus == LedgerTransactionStatus.Succeeded).Select(lt => lt.StateVersion),
```
dhedey, line 193: Out of interest, do we have indices specific to success state success/fail? 

My slight worry is that if not, we might hit the heap which might be slow because each ledger transaction row is massive.

dhedey (approved):

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/TransactionTypeLedgerTransactionMarkerProcessor.cs
PawelPawelec-RDX: nope, it's not possible to exit from function when you're inside switch expression, it has to return some value.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/ResourceBalanceChangeProcessor.cs
PawelPawelec-RDX: yes it's guaranteed, that's basic gateway concept it'd mean we have to comment all places where we use that (50+ places). I'd prefer not to add it.

src/RadixDlt.NetworkGateway.PostgresIntegration/Migrations/IdempotentApplyMigrations.sql
PawelPawelec-RDX: good catch. done.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/TransactionQuerier.cs
PawelPawelec-RDX: you're right we need that index. fixed.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/ResourceBalanceChangeProcessor.cs
dhedey: Well what I mean is - do we explicitly load them from the `FungibleEntityBalanceChanges` etc?
I assume not? So it's implicit that they're loaded in (via some other registrations of these addresses), and a refactor which removed those could theoretically break this. I guess if we have tests, we'll catch this.
When I did Olympia; each processor had a pre-process step (where e.g. it would register things to load from the DB) and then a process step, and it was expected that it would keep these in parallel. Here it's implicit. I think that's okay.
Happy for us not to document if you think it's sufficiently safe/obvious.

src/RadixDlt.NetworkGateway.PostgresIntegration/Services/TransactionQuerier.cs
PawelPawelec-RDX: i did run that simple query on prod and it was looking good, plan was to execute regular performance tests on dummunet with Les.

src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/Processors/LedgerTransactionMarkers/ResourceBalanceChangeProcessor.cs
PawelPawelec-RDX: yes we load them explicitly here by using `IEntityAddressPointer` interface. All addresses returned by that interface are fetched from db. https://github.com/radixdlt/babylon-gateway/blob/main/src/RadixDlt.CoreApiSdk/Model/CommittedTransactionBalanceChanges.cs

dhedey: Ahh nice! 👍

PR #821, opened 2025-01-13
fixes after review.

PawelPawelec-RDX:


src/RadixDlt.NetworkGateway.PostgresIntegration/CommonDbContext.cs
```diff
@@ -651,14 +651,17 @@ private static void HookupHistory(ModelBuilder modelBuilder)
...
             .Entity<ResourceHolder>()
             .HasIndex(e => new { e.EntityId, e.ResourceEntityId })
             .IsUnique();
 
+        // Used by Gateway API when querying resource holders endpoint.
         modelBuilder
             .Entity<ResourceHolder>()
-            .HasIndex(e => new { e.ResourceEntityId, e.Balance, e.Id });
+            .HasIndex(e => new { e.ResourceEntityId, e.Balance, e.EntityId })
+            .IsDescending(true, true, false);
```
dhedey, line 664: Rather minor (but with a very small perf impact on write) why is `ResourceEntityId` descending rather than ascending?

dhedey (approved):

src/RadixDlt.NetworkGateway.PostgresIntegration/CommonDbContext.cs
PawelPawelec-RDX: ah, good catch, it should be (false, true, true)

